PR #4, opened 2012-08-07
Display a better error if Image Factory is unreachable.

n1zyy:
Previously, we would display an error suggesting that there was a
problem and you should check the Factory log. This was misleading
if Factory couldn't be reached, since there would be nothing
in the log.

This still isn't perfect; if Factory is entirely down, it seems
that aeolus-image-rubygem reports a bogus error. But this change
keeps Conductor from introducing any lies of its own about the cause
of the error.


jprovaznik:
works fine, ACK

PR #5, opened 2012-08-13
Adds an additional check in instance_errors to avoid obscure bug

n1zyy:
In certain cases, we somehow manage to end up with an instance variable
that responds to .errors, but doesn't return an array as expected, so
calling .errors.size on it raises an exception. This patch ensures that
the errors respond to a .size method before we call it, since raising an
exception from within code to handle exceptions is not a good thing.
This patch does not address the root cause of this problem.

### DIFF
--- a/src/app/controllers/application_controller.rb
+++ b/src/app/controllers/application_controller.rb
@@ -172,7 +172,7 @@ class ApplicationController < ActionController::Base
     arr = Array.new
     instance_variables.each do |ivar|
       val = instance_variable_get(ivar)
-      if val && val.respond_to?(:errors) && val.errors.size > 0
+      if val && val.respond_to?(:errors) && val.errors.respond_to(:size) && val.errors.size > 0
         hash[:object] = ivar[1, ivar.size]
         hash[:errors] ||= []
         val.errors.each {|key,msg|


n1zyy:
For reasons that don't make a whole bunch of sense to me yet, the place I'm seeing the failure is with an empty instance of an OrderedHash. (`ivar` is `@_config`, and `val.inspect` returns `#<OrderedHash {}>`.)

The problem seems to be that it's actually an `ActiveSupport::Configurable::Configuration`, which will `respond_to` anything and return nil:

```
1.8.7-p357 :006 > foo = ActiveSupport::Configurable::Configuration.new
 => #<OrderedHash {}> 
1.8.7-p357 :010 > foo.respond_to?(:bacon_cheeseburger)
 => true 
```

I'm still not positive where this is actually coming from, but this does make me think that the additional check I add here is necessary.

jistr:
Yeah, when something uses duck typing as boldly as #instance_errors does, I think it's a good practice to make it defensive.

I'll run specs and cucumbers just to make sure and then follow up on this.

jistr:
Okay, everything passes, so ACK and merge. Good work :)

PR #6, opened 2012-08-13
Fix MetadataObject.lookup to handle lazy-loaded classes (BZ#847373)

jeckersb:
https://bugzilla.redhat.com/show_bug.cgi?id=847373

Under the circumstances in the above bug (running rake
dc:create_user), the Pool class (and others) have not been loaded yet
when running MetadataObject.lookup.  Because of this, Pool is not
listed in ActiveRecord::Base.subclasses, and 'klass' ends up being
nil.  When using const_get instead to obtain a handle to the desired
object class, the const_missing code paths run and Pool gets lazy
loaded.


eggmaster:
Ran aeolus-configure with this patch applied and it appears to fix https://bugzilla.redhat.com/show_bug.cgi?id=847373

PR #7, opened 2012-08-14
Add destroy_users_by_pattern task (BZ#847399)

jeckersb:
https://bugzilla.redhat.com/show_bug.cgi?id=847399

To be used by configure to ensure all temporary admins are destroyed.

### DIFF
--- a/src/lib/tasks/dc_tasks.rake
+++ b/src/lib/tasks/dc_tasks.rake
@@ -50,6 +50,29 @@ namespace :dc do
     end
   end
 
+  desc 'Destroy users matching a pattern'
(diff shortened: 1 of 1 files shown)


rwsu:
I was able to delete some rogue temporary users using this command.

rake dc:destroy_users_by_pattern['%temp%']
Using gem require instead of bundler
User temporary-administrative-user-d6896a8091e1b07cb50ba92bab61fbb06a997526270292fc destroyed
User temporary-administrative-user-167227b0b3b594ea0477c8ac21aefa8f720dd730abd83bfa destroyed

Ack.

PR #8, opened 2012-08-16
Adds missing route to show ProviderAccount

petrblaho:
This commit add missing route to /provider_accounts/:id

That route is used for creating links on Account page (/account - Users#show) in permission table.

See http://i.imgur.com/OUK59.png

### DIFF
--- a/src/config/routes.rb
+++ b/src/config/routes.rb
@@ -158,7 +158,7 @@ Conductor::Application.routes.draw do
     end
   end
 
-  resources :provider_accounts, :only => :index
+  resources :provider_accounts, :only => [ :index, :show ]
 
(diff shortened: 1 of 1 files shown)


rwsu:
Nice catch. 

Applying the patch didn't fix the error. It could be a different issue because I got a slightly different error. 

The link to the Provider Account resource points to http://localhost/providers/id/provider_accounts/id. It should be http://localhost/conductor/provider/id/provider_account/id. The "conductor" part is missing.

I see a Apache error page indicating Not Found.

I think we need to also add a cucmber test to exercise this link.

martinpovolny:
@petrblaho: will you update the pull-request?

PR #15, opened 2012-08-24
Basic support for Kerberos SSO authentication

mmahut:
Basic support for web interface SSO authentication with kerberos. It
assumes the HTTP_X_FORWARDED_USER header is passed from the proxy, usually
apache or other web server, in some cases hardware proxy instance with Kerberos
support. After successful verification of HTTP_X_FORWARDED_USER header, this behaves the
same way as ldap. The kerberos authentication is done on the proxy layer due to
security concerns, as the kerberos keytab is often shared among services such as ssh.

Note this is not intended to replace the auth between components, but just to enable
the benefits of SSO as a login into the web interface. There is room for future improvement
too. Maybe even integration with FreeIPA directly into the web interface.

### DIFF
--- a/src/app/models/user.rb
+++ b/src/app/models/user.rb
@@ -127,6 +127,14 @@ class User < ActiveRecord::Base
     return u
   end
 
+  def self.authenticate_using_krb(username, ipaddress)
+    u = User.find_by_login(username) || create_krb_user!(username)
+    u.login_count += 1
+    update_login_attributes(u, ipaddress)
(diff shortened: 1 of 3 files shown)


jprovaznik:
Hi,
the patch works fine, going to merge this.
Only nit (which we already discussed) is with conductor login form, but this is described in this RFE ticket: https://bugzilla.redhat.com/show_bug.cgi?id=851514.

Thanks for this patch.

PR #17, opened 2012-08-24
Pool's default enabled value is true

jtomasek:


n1zyy:
ACK, this looks good! Will merge.

PR #18, opened 2012-08-24
RM3523 Quickfixes - On deployment overview display quota information onl...

jtomasek:
...y when relevant

To add html tags to i18n message variables, the yaml key has to end with "_html"
http://stackoverflow.com/questions/5367705/rails-3-i18n-interpolating-html-tags-into-text


n1zyy:
Looks good, tests good. ACK!

PR #19, opened 2012-08-28
Configurable session timeout + fixed specs

jistr:
- 1st commit: Make session expiration timeout configurable. The timeout is settable in settings.yml.
- 2nd commit: Fix session expiration specs. There was a typo in `before` block which should cause them to fail, but there was another error which caused them to pass anyway. One more spec added as a bonus.

### DIFF
--- a/src/app/controllers/application_controller.rb
+++ b/src/app/controllers/application_controller.rb
@@ -468,7 +468,7 @@ class ApplicationController < ActionController::Base
       # compare to last non-backbone reqest time
       last_time = session[:last_active_request]
       if last_time
-        logout if last_time < 15.minutes.ago
+        logout if last_time < SETTINGS_CONFIG[:session][:timeout].minutes.ago
       else
         # This isn't really an active reqest, but
         # boostrapping session time if none set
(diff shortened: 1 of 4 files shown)


jistr:
Also, there's something in current check_session filter behavior that I believe might be a bug. The check_session code allows for this:
- Let's assume that session expiration timeout is 15 minutes.
- Make a normal request.
- Wait 14 minutes.
- Make a Backbone request.
- Wait 14 minutes.
- Make a normal request. _This request will not have an expired session. There will be 28 minutes between two normal requests even though the timeout is set to 15 minutes._

In other words, if this request [1] would be a normal non-backbone request, the spec would fail.

Do we consider this a bug? I have a fix almost ready and can add it to this pull rq.

[1] https://github.com/jistr/conductor/commit/83266ced53d8d34435a447f4209c824f31ed4175#L0R45

jguiditta:
That sounds like a bug, yes.  I would suggest making it a separate pull request though to keep this one focused on the  new functionality of making session timeout configurable.

jguiditta:
Ok, this all seems to work fine, good work! ACK and pushed to master

PR #20, opened 2012-08-29
Removed the background, ...

jtomasek:
... replaced with gradient that maintains the color and visibility of header border and footer background.


mtaylor:
Careful with your use of words on this public list.  We don't want a 
$660 million course case to pay for :P

n1zyy:
Revisiting this, I'm going to ACK+merge and send a separate pull request to remove the file. No sense in delaying this a day.

PR #21, opened 2012-08-29
Bz816054 fix quota error msg

ifarkas:
Since the whole provider selection changed, the issue is not reproducible as it is stated in the BZ ticket. However, some improvements need to be made for the notifications.

This patchset also implements Jan's suggestion to refactor the deployment#find_match_with_common_account method into 2 seperate methods.

https://bugzilla.redhat.com/show_bug.cgi?id=816054

### DIFF
--- a/src/app/models/deployment.rb
+++ b/src/app/models/deployment.rb
@@ -234,7 +234,7 @@ class Deployment < ActiveRecord::Base
     self.state = STATE_PENDING
     save!
 
-    all_inst_match, account, errs = find_match_with_common_account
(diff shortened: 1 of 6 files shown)


n1zyy:
I did run tests and they looked clean. The code also looks good. I have one minor nit, and then some questions (visible above) about the string changes. Angus has been cleaning up strings, so I'd like to check with him in the morning to make sure the strings are perfect.

n1zyy:
Okay, this looks good and fixes the problem I had noted with strings. ACK, and merging!

PR #24, opened 2012-08-30
Travis-friendly cukes

jistr:
Green cukes when factory/iwhd/deltacloud services are down.

With this, cucumber should be green on Travis.

### DIFF
--- a/src/features/step_definitions/pool_family_steps.rb
+++ b/src/features/step_definitions/pool_family_steps.rb
@@ -26,13 +26,13 @@ end
 
 Given /^there is a pool family named "([^\"]*)"$/ do |name|
   @pool_family = FactoryGirl.create(:pool_family, :name => name)
-  step %{there are no images in "#{@pool_family.name}" pool family}
+  Given %{there are no images in "#{@pool_family.name}" pool family}
 end
 
 Given /^there is a pool family named "([^\"]*)" with a pool named "([^\"]*)"$/ do |pool_family, pool|
   @pool_family = FactoryGirl.create(:pool_family, :name => pool_family)
   @pool = FactoryGirl.create(:pool, :name => pool, :pool_family => @pool_family)
-  step %{there are no images in "#{@pool_family.name}" pool family}
+  Given %{there are no images in "#{@pool_family.name}" pool family}
 end
 
 Given /^there are no images in "([^\"]*)" pool family$/ do |name|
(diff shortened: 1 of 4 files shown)


src/features/step_definitions/pool_family_steps.rb
```diff
@@ -26,11 +26,17 @@
 
 Given /^there is a pool family named "([^\"]*)"$/ do |name|
   @pool_family = FactoryGirl.create(:pool_family, :name => name)
+  step %{there are no images in "#{@pool_family.name}" pool family}
```
jguiditta: Unfortunately, while 'step' is nice and clean, and this does make tests pass upstream, it also causes things to fails running, for example on fedora 16 with system ruby/rpmified gems.  This appears to be because the step method did not exist in the older version.  Much as I want cukes passing upstream, I do not want it to be at the price of them passing for our platform/distro users.

jguiditta:
https://github.com/cucumber/cucumber/blob/master/History.md#v112 looks like when it was added., fedora 16 has rubygem-cucumber-1.0.1-1.fc16.noarch

jistr:
Oh, that's how it ends up when one thinks it's too small a change to run tests on both VMs... :)

I'll do the aeolus-configure patch related to configurable timeout and then get back to this.

Also, I talked with @petrblaho this morning about the possibility of setting up Travis to check pull requests for us. Travis then comments on each one whether the tests pass. [1]

Travis has a thing called "Build Matrix", allowing us to test against multiple gemfiles [2], so maybe instead of running
- 1.9.3 + latest gems ; 1.8.7 + latest gems

we could run:
- 1.9.3 + latest gems ; 1.8.7 + legacy (F16) gems. We would need to create F16-specific Gemfile for this.

We can discuss this further on IRC.

[1] https://github.com/petrblaho/blog-test/pull/1#issuecomment-8185959
[2] at the bottom of the page: http://about.travis-ci.org/docs/user/build-configuration/

jistr:
Now it passes on both F17 + upstream and F16 + rpms. However I cannot rebase to current master or I'd have to force push :)

jguiditta:
I don't know what I am doing here, but I cannot seem to get more that the 2 commits you originally put in this request. Any chance you can just send this to list, or create a new pull request?

jguiditta:
NVM, pebkac, tests running now, I should be able to push this in a bit

jguiditta:
Ok, this runs against both system and upstream dependencies with services off, woohoo! I have pushed it to master, though I squashed the new commit into the original, so you will only see 2 patches on master.

jistr:
Thanks for reviewing my pull requests :)

PR #26, opened 2012-08-31
BZ811890 Remove untranslated title attributes from Pool Family and...

jtomasek:
... Realm forms

https://bugzilla.redhat.com/show_bug.cgi?id=811890


n1zyy:
ACK, looks good.

PR #27, opened 2012-08-31
BZ 843932 username display

n1zyy:
This fixes https://bugzilla.redhat.com/show_bug.cgi?id=843932 - "UI is quirky with users with no real name"

When using LDAP, it is possible to log in as a user who has no real name specified in LDAP. In Conductor, we then show your name as a blank space, which is broken and looks awful. This patch causes us to fall back to the user's login name if the username is blank.

I added a helper method for this, but I also wanted to use the same logic in some controllers, which caused a little bit of duplication. It would have been a lot simpler to add this to the model, so that `user.name` would do this. However, I think that would be bad, because sometimes when we call `user.name` we do indeed want the user's real name only, like in a table where we show the user's login name in one column and the user's real name in another. Thus, this.

Note that if you enable LDAP authentication to test this, https://www.aeolusproject.org/redmine/issues/3780 will fail, which is seemingly unrelated to this patch.

### DIFF
--- a/src/app/controllers/users_controller.rb
+++ b/src/app/controllers/users_controller.rb
@@ -68,7 +68,7 @@ class UsersController < ApplicationController
   def show
     @user = params[:id] ? User.find(params[:id]) : current_user
     require_privilege(Privilege::VIEW, User) unless current_user == @user
-    @title = @user.name
+    @title = @user.name.present? ? @user.name : @user.login
     @quota_resources = @user.quota.quota_resources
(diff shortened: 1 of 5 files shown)


bmclaughlin:
On 08/31/2012 02:05 PM, Matt Wagner wrote:

> git pull https://github.com/n1zyy/conductor bz_843932-username-display
> Ack, works as advertised.

Brian

PR #28, opened 2012-08-31
Bz 852531 login redirect

n1zyy:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=852531

This ended up being a really easy fix, though I set out trying to solve a challenging problem and thus spent forever walking through session code.

It seems like `redirect_back_or_default` is meant to redirect the user to redirect to `session[:return_to]`, or otherwise the default. But it was actually doing the opposite -- if you passed it a URL, it would redirect there. Otherwise, it would redirect back.

That has to be a bug, because if you wanted to override where it redirected you to (as we do when logging out), you should just redirect them normally.

I added a quick Cucumber test for this.

### DIFF
--- a/src/app/controllers/application_controller.rb
+++ b/src/app/controllers/application_controller.rb
@@ -235,7 +235,7 @@ class ApplicationController < ActionController::Base
   end
 
   def redirect_back_or_default(default)
-    redirect_to(default || session[:return_to])
+    redirect_to(session[:return_to] || default)
     session[:return_to] = nil
   end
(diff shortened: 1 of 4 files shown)


n1zyy:
One thing to note here -- this seems like a recent regression, but the `redirect_back_or_default` method was last changed in 2010, so I still don't understand what caused it to stop working recently. My patch fixes the bug though.

jtomasek:
ACK

PR #34, opened 2012-09-04
BZ #854145: limit the time frame for the penalty for failure strategy

ifarkas:
https://bugzilla.redhat.com/show_bug.cgi?id=854145


petrblaho:
ACK, works for me.

ifarkas:
Pushed. Thanks for the review!

PR #35, opened 2012-09-04
BZ851107 New deployment form validation fix

jtomasek:
The Deployment is now validated after clicking next and proper error message is displayed if validation fails.
Next button is no more disabled. Js only evaluates if the name is available and displays message next to input.

https://bugzilla.redhat.com/show_bug.cgi?id=851107


ifarkas:
ACK with 1 inline comment

jtomasek:
Updated and pushed to master and 1.1

PR #36, opened 2012-09-04
BZ 783128 - Renames "Image URI" to "Provider's Image ID"

n1zyy:
Beyond the fact that it's not actually a URI at all, it's unclear
what it is. We decided "Provider's Image ID" was clearer.

https://bugzilla.redhat.com/show_bug.cgi?id=783128

### DIFF
--- a/src/app/controllers/deployments_controller.rb
+++ b/src/app/controllers/deployments_controller.rb
@@ -256,53 +256,84 @@ class DeploymentsController < ApplicationController
 
   def destroy
     deployment = Deployment.find(params[:id])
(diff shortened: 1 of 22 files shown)


jtomasek:
ACK, just a note: I think the key for the translation should be `image_id` instead of `image_uri` (see line 918 in en.yml). I would even prefer using `providers_image_id: "Provider's Image ID"` for this translation as `image_id` translates to "Image ID" in other places in application.

n1zyy:
Well that didn't go well. I'm going to close this and open a new one rather than try to do something fancy with force-pushing like I just did...

PR #38, opened 2012-09-04
BZ 790817 - Tooltips for Pool scoreboard

n1zyy:
https://bugzilla.redhat.com/show_bug.cgi?id=790817


jtomasek:
ACK, and pushed. I will create patch for the pool detail scoreboard.

PR #39, opened 2012-09-05
Quota consistency

rwsu:
https://www.aeolusproject.org/redmine/issues/3758

Changed show to display maximum_running_instances as an attribute.

Examples

<quota maximum_running_instances='100'></quota>
<quota maximum_running_instances='unlimited'></quota>

Noticed provider accounts doesn't test quota. We may have to revisit how quota is specified as xml for provider_accounts_controller.rb


petrblaho:
ACK, PUSHED.

PR #41, opened 2012-09-05
BZ #810957: Mustachifying images#index

ifarkas:
https://bugzilla.redhat.com/show_bug.cgi?id=810957


jtomasek:
ACK, the image list gets updated automatically when new image is added.

ifarkas:
Pushed

PR #45, opened 2012-09-05
Adds QuotaAware to Provider Account create and update

petrblaho:
This changes handling of quota params in ProviderAccountsController for create and update actions to use QuotaAware module. This ensures that quotas are handled consistently and XML input uses the similar format.


rwsu:
Ack.

PR #47, opened 2012-09-06
Refactor count_uptime helper

jtomasek:
This patch changes the count_uptime helper to use time_ago_in_words


ifarkas:
This looks sane and I prefer these user-friendly time messages which are provided by this patch, but there isn't enough space for them in the pretty view boxes on pools#index. Since the boxes are about to be redesigned we can wait with this patch until it's done.

I also added 1 inline note.

So delayed ACK :-)

martinpovolny:
Broken :-(

PR #48, opened 2012-09-06
Bz848472

petrblaho:
This is solution for https://bugzilla.redhat.com/show_bug.cgi?id=848472

Patch contains correct parsing of provider_type_id form params, new validation for missing provider_type in Provider, removes default value from provider_type_id column in providers table and correcting a type in providers_controller_spec.rb


petrblaho:
I created rev2 patch for this - https://github.com/aeolusproject/conductor/pull/49

PR #50, opened 2012-09-10
Converge ui update

jtomasek:
This patchset updates conductor to the latest converge-ui.
- Updated to latest converge-ui
- Renamed user.login to user.username
- Layout changes, login page uses converge-ui layouts, new style for notifications, login works with or without javascript
- Removed unused layout and partials
- Updated spec file (converge-ui-devel is new BuildRequires dependecy)

(don't forget to git submodule update after applying this patch)


n1zyy:
I need to dig deeper, but I'm having trouble running this on upstream Ruby+bundler. Logging in leaves me with:

```
"undefined local variable or method `view_context' for #<#<Class:0x7f5a8438f528>:0x7f5a851b8168>"
```

That error is not present on system Ruby.

n1zyy:
Okay, as an update: that error above is _not_ introduced by this patch. It was introduced to master by c3b1d79b729d01fa0b91455510466f3871247b29 and just came along when I updated to run this patch.

jtomasek:
I have found an error in spec file. I will create new pull request rebased to conductor master. Separate patch for aeolus-configure is also needed, I have that one almost ready.

jtomasek:
I have made fixed, rebased and more up to date pull request for this: https://github.com/aeolusproject/conductor/pull/52

PR #53, opened 2012-09-11
BZ #855717: fix Ruby 1.8 compatibility

ifarkas:
https://bugzilla.redhat.com/show_bug.cgi?id=855717

In Ruby 1.8 the Object#methods returns an array of string, while in Ruby 1.9 it returns an array of symbols.
Rails provides Object#method_names which always returns an array of string.


tzumainn:
ACK

tzumainn:
ACK'd and merged.

Mainn

----- Original Message -----

PR #54, opened 2012-09-11
Refactor Ruby 1.8 compatibility fix in provider selection

ifarkas:
The change was introduced by 7f26b8846b36515a6648b3f94ab24036471d8370 which is using a Rails method to solve an incompatility issue between 1.8 and 1.9.
It's better no to couple Rails with the provider selection because in the future we may want to seperate the provider selection code into a separate gem.


tzumainn:
Tests pass and original problem is still fixed, so ACK.

PR #55, opened 2012-09-11
BZ 760592 (rev.2) - use OAuth Client instead of Restclient for oauth

blomquisg:
https://bugzilla.redhat.com/show_bug.cgi?id=760592

Switch conductor to use the ruby oauth client instead of restclient when
connecting to config server.  With restclient, the post body is not properly
signed.

Rev2:  Fixed cucumber tests:  Removed "require_relative" from cukes and copied over "FakeResponse" from specs into cuke's support.


tzumainn:
Verified that config servers work and that tests pass.  ACK.

PR #56, opened 2012-09-12
Removed unused content_fors from user_sessions/new ...

jtomasek:
... which are not used in converge-ui any more


ifarkas:
ACK

jtomasek:
Thanks, pushed to master.

PR #57, opened 2012-09-12
Quick header button fixes

n1zyy:
This is a pair of really minor patches to fix up the buttons in our header.

They had an inexplicable gray border, and were missing the `header-widget` class that Katello used.


ifarkas:
ACK

PR #58, opened 2012-09-13
updated converge-ui to 1.0 version, removed unnecessary header styles,

jtomasek:
updated requirement for converge-ui-devel 1.0.3 on spec.in

This patch updates conductor to work with the converge-ui converge-ui-devel-1.0.3-1, origin/CONVERGE-UI-1.0 (commit 26ce309)

This patch should go to both master and 1.1 branches.

CONVERGE-UI-1.0 is the point from which converge-ui-devel should be built (for conductor 1.1)

### DIFF
--- a/aeolus-conductor.spec.in
+++ b/aeolus-conductor.spec.in
@@ -74,7 +74,7 @@ Requires: curl
 BuildRequires: rubygem(sass)
 BuildRequires: rubygem(compass) >= 0.11.5
 BuildRequires: rubygem(compass-960-plugin) >= 0.10.4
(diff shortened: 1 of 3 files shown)


n1zyy:
This patch doesn't add the html5shiv stuff; that didn't make it into the "CONVERGE-UI-1.0" branch you point to. That's fine, but it's not what the commit message says.

n1zyy:
This looks great. GitHub apparently allows me to edit other peoples' commit messages in pull requests (!), so now this is looking perfect. Will merge and pull over to 1.1!

PR #60, opened 2012-09-13
Removes a malformed haml comment from layout

n1zyy:
This was displaying the haml as an HTML comment, which was
kind of silly. Since it wasn't used, I'm just removing it.


n1zyy:
https://bugzilla.redhat.com/show_bug.cgi?id=857223

ifarkas:
ACK

It would be better to include the BZ link in the commit message, so I am not going to merge this pull request. Could you  add it and then merge?

n1zyy:
Good idea, Imre, will do.

PR #64, opened 2012-09-17
BZ 858030 - Allow 0 storage in HWPs

n1zyy:
A hardware profile with no storage is a valid configuration,
so Conductor should accept it.

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=858030


jprovaznik:
Looks good and works fine.

PR #71, opened 2012-09-21

rwsu:

API docs updated at https://www.aeolusproject.org/redmine/projects/aeolus/wiki/Pool_Families

I am curious as to people's opinion on how to do these sorts of mappings. As implemented a user passes in provider ids as a list of options in an accounts_selected array. Is there a better way of doing this?

### DIFF
--- a/src/app/controllers/pool_families_controller.rb
+++ b/src/app/controllers/pool_families_controller.rb
@@ -149,7 +149,10 @@ class PoolFamiliesController < ApplicationController
 
     if ProviderAccount.count == 0
       flash[:error] = t('pool_families.flash.error.no_provider_accounts')
-      redirect_to pool_family_path(@pool_family, :details_tab => 'provider_accounts') and return
+      respond_to do |format|
+        format.html { redirect_to pool_family_path(@pool_family, :details_tab => 'provider_accounts') and return }
+        format.xml { raise(Aeolus::Conductor::API::Error.new(500, flash[:error])) }
+      end
     end
 
     @provider_accounts = ProviderAccount.
@@ -163,6 +166,10 @@ class PoolFamiliesController < ApplicationController
 
     if params[:accounts_selected].blank?
       flash[:error] = t"pool_families.flash.error.select_to_add_accounts" if request.post?
+      respond_to do |format|
+        format.html
+        format.xml { raise(Aeolus::Conductor::API::Error.new(500, flash[:error])) }
+      end
     else
       ProviderAccount.find(params[:accounts_selected]).each do |provider_account|
         if check_privilege(Privilege::USE, provider_account) and
@@ -182,6 +189,13 @@ class PoolFamiliesController < ApplicationController
       respond_to do |format|
         format.html { redirect_to pool_family_path(@pool_family, :details_tab => 'provider_accounts') }
         format.js { render :partial => 'provider_accounts' }
+        format.xml {
+          if flash[:error]
+            raise(Aeolus::Conductor::API::Error.new(500, flash[:error]))
+          else
+            render :show, :locals => { :pool_family => @pool_family }
+          end
+        }
       end
     end
   end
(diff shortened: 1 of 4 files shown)


petrblaho:
On Thu, Sep 20, 2012 at 05:28:49PM -0700, Richard Su wrote:

> https://www.aeolusproject.org/redmine/issues/3850
> 
> API docs updated at https://www.aeolusproject.org/redmine/projects/aeolus/wiki/
> Pool_Families
> 
> I am curious as to people's opinion on how to do these sorts of mappings. As
> implemented a user passes in provider ids as a list of options in an
> accounts_selected array. Is there a better way of doing this?
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> You can merge this Pull Request by running:
> 
>   git pull https://github.com/rwsu/conductor poolfam-provideraccount
> 
> Or view, comment on, or merge it at:
> 
>   https://github.com/aeolusproject/conductor/pull/71
> 
> Commit Summary
> 
>   • RM 3850 - show, add, and remove provider accounts within a pool family
> 
> File Changes
> 
>   • M src/app/controllers/pool_families_controller.rb (27)
>   • M src/app/views/pool_families/_detail.xml.haml (5)
>   • M src/config/routes.rb (8)
>   • M src/spec/controllers/pool_families_controller_spec.rb (84)
> 
> Patch Links
> 
>   • https://github.com/aeolusproject/conductor/pull/71.patch
>   • https://github.com/aeolusproject/conductor/pull/71.diff
> 
> —
> Reply to this email directly or view it on GitHub.
> 
> *

Hi,

creating associations between resources is interesting issue.

I do not like /api/pool_families/:pool_family_id/add_provider_accounts url at all.
It sounds like verb and not like noun - it does not sound like resource.

Discussions with @mtaylor and @jistr about how API should
be built led us to agree that we should design API around resources -
nouns.

So I would propose to think about this association as a resource - what
it is? It is a link between pool_family and provider_account.

Me and @jistr talked about possibilities how that part of API can look.

<pre>/api/pool_families/:pool_family_id/provider_accounts</pre> route answering to GET method
and returning list of provider_accounts linked with specified (by
:pool_family_id in url) pool_family.
Next there is
<pre>/api/pool_families/:pool_family_id/provider_accounts/:provider_account_id</pre>

and it should answer to PUT (create link between two), DELETE (destroy
link between two) and GET (get info whether the link exists or not).

This urls contains "provider_accounts" part but that does not mean it is the resource
provider_account itself. I failed to come out with better name for that
link between pool_family and provider_account.

The whole idea behind this is that:
1) we work with resources, not commands
2) we are creating link between 2 resources, not updating resources
itself. I think that this link should be treated as resource itself

I would like to read other people's opinions on this thing.
I hope we can find good solution for our project.

## 

With regards
Petr Blaho

jistr:
I agree with @petrblaho, we already discussed this in person.

The only thing that remains unclear here is, as @petrblaho pointed out, the name of the nested route:

> This urls contains "provider_accounts" part but that does not mean it is the resource
> provider_account itself. I failed to come out with better name for that
> link between pool_family and provider_account.

Personally, I think naming it `provider_accounts` is ok. Name like `provider_accounts_associations` is maybe more precise but is kinda long and I don't think it brings any benefit. I don't think we will ever want to have full subresource like `provider_accounts` and an association subresource like `provider_accounts_associations` at the same time, so I would simply name the association subresource `provider_accounts`.

BTW I checked again how GitHub does that, and they use **exactly** the same approach that @petrblaho just proposed to for creating/deleting associations.
- `GET /resources/id_1/subresources` to list all associations
- `GET /resources/id_1/subresources/id_2` to check whether an association exists
- `PUT /resources/id_1/subresources/id_2` to create an association
- `DELETE /resources/id_1/subresources/id_2` to delete an association

http://developer.github.com/v3/orgs/teams/#list-team-members

One thing to note, that I like very much, is that e.g. `GET /resources/id_1/subresources/id_2` returns 204 No Content and has no response content when the association exists. This makes it apparent that you shouldn't mistakenly use the association route when you want to get full XML representation of the subresource. It doesn't return any XML. It only informs you about the association status by returning 204 or 404. I'd say we go that way as well.

Another thing to note here is that the name of association route does not necessarily have to be the same as the name of the route for resource it represents. In GitHub's case the association route is `/teams/id_1/members/id_2`, but when you want details about the member, you query `/users/id`, not `/members/id`. This is probably obvious, so forgive me for pointing it out. I just wanted to avoid any confusion, because in case of `provider_accounts` the two routes probably make most sense to be named the same, but it isn't always so.

rwsu:
Creating a resource to manage these association does make sense. I will look into refactoring this.

Thanks for the excellent feedback.

PR #72, opened 2012-09-21
BZ 858997 - Fix handling of unauthenticated XHR requests

n1zyy:
Both the login page and the rest of the app now support handling
an HTTP 401 response when the request is unauthenticated.

https://bugzilla.redhat.com/show_bug.cgi?id=858997

### DIFF
--- a/src/app/views/user_sessions/new.html.haml
+++ b/src/app/views/user_sessions/new.html.haml
@@ -10,6 +10,11 @@
   = javascript_include_tag "converge-ui/vendor/jquery/jquery-1.6.2.js"
   = javascript_include_tag "converge-ui/vendor/rails.js"
   = javascript_include_tag "converge-ui/login.js"
+  :javascript
+    $("#login_form").bind("ajax:error", function(event, data, status, xhr) {
+      $("#notifications").html("<div class=\'error\'>\n <p class=\'text\'>#{t("user_sessions.flash.warning.login_failed")}<\/p>\n<\/div>\n");
+      $('form .spinner').fadeOut('fast');
+    });
   /[if lt IE 9]
 
 = content_for :title do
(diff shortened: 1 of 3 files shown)


src/app/views/user_sessions/unauthenticated.js.erb
```diff
@@ -1 +0,0 @@
-$("#notifications").html("<%= escape_javascript(render :partial => "notifications") %>");
```
jtomasek, line 1: This file could get deleted as it is empty.

jtomasek:
ACK, I think this ended up as quite nice solution. Works as expected. Some nits inline.

src/app/views/user_sessions/new.html.haml
```diff
@@ -10,6 +10,11 @@
   = javascript_include_tag "converge-ui/vendor/jquery/jquery-1.6.2.js"
   = javascript_include_tag "converge-ui/vendor/rails.js"
   = javascript_include_tag "converge-ui/login.js"
+  :javascript
+    $("#login_form").bind("ajax:error", function(event, data, status, xhr) {
+      $("#notifications").html("<div class=\'error\'>\n <p class=\'text\'>#{t("user_sessions.flash.warning.login_failed")}<\/p>\n<\/div>\n");
+      $('form .spinner').fadeOut('fast');
+    });
```
jtomasek, line 19: This should be surrounded by `$(document).ready(function(){ ... });` just to be safe.

src/app/views/user_sessions/unauthenticated.js.erb
n1zyy, line 1: I believe it is deleted and GitHub is just not showing it clearly. The file no longer exists in the branch I pushed.

jtomasek:
ACK

PR #74, opened 2012-09-24
BZ #859503: laucn new deployment button on deployables#show redirects to deployments#launch_new instead of deployments#launch_time_params

ifarkas:
https://bugzilla.redhat.com/show_bug.cgi?id=859503


tzumainn:
ACK

PR #75, opened 2012-09-24
BZ859926 Deployments preset filters translation fix

jtomasek:
This patch corrects the translation key for deployments preset filter "All deployments" option.

https://bugzilla.redhat.com/show_bug.cgi?id=859926


jprovaznik:
ACK, works for me

PR #79, opened 2012-09-27
BZ 751088 (refix): Configserver and conductor on same box

blomquisg:
https://bugzilla.redhat.com/show_bug.cgi?id=751088

This patch changes the way conductor builds the request paths when
talking with configserver.  Previously, conductor assumed that
configserver ran at the root context of the server (i.e.,
"http://configserver/").  Now, conductor can adapt to a configserver
that runs at a non-root context (i.e., "http://servername/configserver").

Compatible with aeolus-configserver-0.4.11.

### DIFF
--- a/src/app/models/config_server.rb
+++ b/src/app/models/config_server.rb
@@ -29,6 +29,7 @@
 
 
 require 'cgi'
+require 'uri'
 require 'oauth'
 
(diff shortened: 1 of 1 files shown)


tzumainn:
ACK

PR #84, opened 2012-10-01
Add an rvmrc file to the project.

jguiditta:


n1zyy:
ACK. Consistency is awesome, and this has no adverse impact on people not using rvm.

PR #86, opened 2012-10-02
Realm renaming

rwsu:
https://www.aeolusproject.org/redmine/issues/3788

Renamed Realm model to ProviderRealm and RealmController to FrontendRealmController.


rwsu:
Scott noted that we may need to update the upgrade script to move existing realms over. This should be investigated before this pull request is merged in.

rwsu:
Updated and posted new pull request.

PR #88, opened 2012-10-02
Catalogs API

jistr:
There are several commits that implement the API. The last commit [1] is a fix for both UI and API to check for permissions on Catalogs when listing Catalogs for a given Pool. (This means using `@pool.catalogs.list_for_user` instead of just `@pool.catalogs`.)

[1] https://github.com/jistr/conductor/commit/81120f25e57800c6f17e5205539c1ad6a450a78b


rwsu:
Ack.

The travis spec test run failed. Judging from the logs, it looks like it was a infrastructure issue. I ran the tests locally for 1.8.7 and they passed.

jistr:
Thanks for the review :)

PR #89, opened 2012-10-02
Bumped deltacloud-client version in Gemfile.lock

jprovaznik:
0.5.0 is quite obsolete now


jguiditta:
We need to use bundler to update the lock file, otherwise we dont get all updated deps that this requires.

PR #90, opened 2012-10-02
BZ 861063 - Post-login redirect goes to referer for non-GET URLs

n1zyy:
The existing redirect code to return people to their previous
URL did not consider request type, and would thus redirect you to
various URLs using a GET which was usually incorrect.

For non-GET URLs, we should instead redirect to the referring page.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=861063

### DIFF
--- a/src/app/controllers/application_controller.rb
+++ b/src/app/controllers/application_controller.rb
@@ -256,7 +256,7 @@ class ApplicationController < ActionController::Base
   end
 
   def store_location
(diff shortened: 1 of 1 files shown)


n1zyy:
Bump -- this has sat for 2 weeks, but still needs review. Anyone willing to take a quick look?

tzumainn:
ACK, merging

PR #95, opened 2012-10-03
Fix backwards logic in Provider.valid_framework?

jeckersb:
Looks like this got reversed in 598c9e9b; it causes adding new
providers to fail.  Just return the truthiness of connect!, since it
returns the client object if successful.


n1zyy:
This Travis failure is erroneous -- it failed because "bundle install" timed out. Tests pass locally, and this fixes the issue. ACK; merging.

PR #96, opened 2012-10-04
BZ 861471 - fixed saving of changed launch params in deployment

jprovaznik:
Deployment's deployable_xml attribute was not updated properly before saving.
Proper fix of this would be updating params in deployable_xml right after they
are changed (instead of doing this in before_save filter), but this patch
might introduce some regressions which is not ideal before 1.2 release.

https://bugzilla.redhat.com/show_bug.cgi?id=861471


n1zyy:
Looks good, merging!

PR #97, opened 2012-10-04
BZ #862295: fix back link on deployments#launch_new for RPM installations

ifarkas:
https://bugzilla.redhat.com/show_bug.cgi?id=862295


n1zyy:
ACK, works on both RPM builds and from source.

PR #98, opened 2012-10-04
Change 'url' attributes to 'href' in APIs

jistr:
We now use 'href' attribute everywhere to reference other API resources
in XML responses.


jguiditta:
This looks good, merging

PR #99, opened 2012-10-04
Correction of using bad model in cukes step defs

petrblaho:
This patch is solution for https://www.aeolusproject.org/redmine/issues/3888

There was bad model used for determining id in cucumber step definitions.
Model used was CatalogEntry instead of Deployable.


jguiditta:
Great, this seems to fix tests so 'rake' now works.  What a weird bug

PR #100, opened 2012-10-05

rwsu:

Refactored https://github.com/aeolusproject/conductor/pull/71 to use
a provider_accounts resource under pool_families to manage the
associations.


jistr:
Conditional ACK with respect to the things mentioned above.

PR #101, opened 2012-10-05
Consistent error messages for API

jistr:
**_Please ping me on IRC before merging.**_

Validation errors follow the structure of other API errors. E.g.:

```
<error>
  <code>ValidationError</code>
  <message>Pool can't be blank. Name can't be blank.</message>
</error>
```

Code for some specs and cucumber steps is updated accordingly.

### DIFF
--- a/src/app/views/api/validation_error.xml.haml
+++ b/src/app/views/api/validation_error.xml.haml
@@ -1,5 +1,5 @@
 !!!XML
-%errors
-  - errors.full_messages.each do |message_error|
-    %error
-      %message= message_error
+%error
+  %code ValidationError
+  -# Rails error messages don't have dots at the end, we need to append them
+  %message= errors.full_messages.map { |message| message + '.' }.join(' ')
--- a/src/features/step_definitions/api_steps.rb
+++ b/src/features/step_definitions/api_steps.rb
@@ -101,10 +101,9 @@ Then /^I should receive(?: an?)? (.+) error$/ do |status_name|
   response.headers['Content-Type'].should include('application/xml')
   response.status.should be_eql(status_codes[status_name])
   xml_body = Nokogiri::XML(response.body)
-  # FIXME the XPath expectation should be more restrictive once we standardize the API
-  # As of 2012-09-24, some actions return <errors> with multiple <error> inside
-  # and others return just <error>
-  xml_body.xpath('//error').size.should >= 1
+  xml_body.xpath('//error').size.should == 1
+  xml_body.xpath('/error/code').size.should == 1
+  xml_body.xpath('/error/message').text.length.should > 0
 end
 
 def send_xml_get(uri)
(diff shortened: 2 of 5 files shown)


src/app/views/api/validation_error.xml.haml
```diff
@@ -1,5 +1,5 @@
 !!!XML
-%errors
-  - errors.full_messages.each do |message_error|
-    %error
-      %message= message_error
+%error
+  %code ValidationError
+  -# Rails error messages don't have dots at the end, we need to append them
+  %message= errors.full_messages.map { |message| message + '.' }.join(' ')
```
mtaylor, line 5: You are iterating through the array twice here: First with map, then with join.

Can you not do the same thing with:

%message= errors.full_messages.join('. ')

This only iterates through the array once.

Also, I would suggest changing the delimiter to something other than ". " It's possible that this could appear one of the messages.  For example at the end of a sentence.  This makes it difficult to parse the individual messages in the client.  How about splitting on newline or semi colon?

Lastly,  I'd suggest using "#{message}."  It's safer than using + (it will not raise exceptions when message is not a string).

mtaylor:
The structure looks good and tests pass in Travis.  Minor comments aside this looks good.  Please update with respect to comments.  Ping me on irc if anything doesn't make sense.

Thanks

Martyn

src/app/views/api/validation_error.xml.haml
jistr, line 5: > Can you not do the same thing with:
> 
> %message= errors.full_messages.join('. ')

Hmm, I don't think so. With this code, the last message won't have a dot at the end of the sentence.

The purpose of adding a dot was to make the messages appear as sentences (because Rails will give you just messages like "Name can't be blank" <-- without dot at the end), and the purpose of the join was to join the sentences together separated by spaces. So in case we want to make it into sentences, we either loop twice or loop once and then append a final dot. The first solution is clearer in code,  the second saves cycles, but I don't think it's going to make any difference. The computational complexity still remains O(n) and the number of errors will be usually < 10, so I'd go with the clearer solution.

> Also, I would suggest changing the delimiter to something other than ". " It's possible that this could appear one of the messages. For example at the end of a sentence. This makes it difficult to parse the individual messages in the client. How about splitting on newline or semi colon?

I'd rather go for semicolon then :)

> Lastly, I'd suggest using "#{message}." It's safer than using + (it will not raise exceptions when message is not a string).

Nice, thanks :)

So I'd make it either this:

```
%message= errors.full_messages.map { |message| "#{message}."}.join(' ')
```

which would return string more friendly to a human:

```
 <message>Pool can't be blank. Name can't be blank.</message>
```

or this:

```
%message= errors.full_messages.join(';')
```

which would return string more friendly to a machine:

```
<message>Pool can't be blank;Name can't be blank</message>
```

jistr, line 5: Actually, let's make it the semicolon solution like you suggested. API is meant to be consumed by machines in the first place :)

I will not put a space after semicolon, so that the client app does not have to strip the strings.

Thanks Martyn, good review ;)

PR #102, opened 2012-10-05
BZ863048 New deployable from image catalog list js fix

jtomasek:
rails translations can't be used in javascript, so the translation strings
are passed using data attribute

fixes https://bugzilla.redhat.com/show_bug.cgi?id=863048


tzumainn:
ACK, merging

PR #103, opened 2012-10-05
clean up some deprecations

martinpovolny:
3 fixes of deprecation warnings

### DIFF
--- a/src/app/models/quota.rb
+++ b/src/app/models/quota.rb
@@ -34,7 +34,7 @@
 
 class Quota < ActiveRecord::Base
 
-  set_table_name "quotas"
+  self.table_name = "quotas"
 
   has_one :pool
   has_one :pool_family
--- a/src/app/models/view_state.rb
+++ b/src/app/models/view_state.rb
@@ -30,7 +30,7 @@
 #
 
 class ViewState < ActiveRecord::Base
-  set_primary_key :uuid
+  self.primary_key = :uuid
   belongs_to :user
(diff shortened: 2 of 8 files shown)


src/Gemfile
```diff
@@ -40,4 +40,5 @@ group :development, :test do
   gem 'vcr', '< 2.0.0'
   gem 'webmock'
   gem 'launchy'
+  gem 'minitest'
```
jguiditta: This is not our dependency, it is somewhere in the rails stack, so we should not have to add this ourselves.  Did you hit this on fedora 17?  If so, it is because of the broken ruby 1.9 they ship, which does not include stdlib, where minitest is

martinpovolny:
sure, it was on fedora 17, shall I remove this change from the pull request?

jtomasek:
Merging this pull request will introduce two unnecessary commits, I think better solution is to reset wrong commit in your local branch (or do the interactive rebase beforehand if needed) and then push --force the local branch. The pull request will then get automatically updated. (Now with only the correct commits)

jguiditta:
@jtomasek this is a good point - we may lose the comment thread is my only concern, but not a huge deal.  I don't like the idea of extraneous commits either, if we can avoid them

martinpovolny:
done as you requested

jguiditta:
Looks good - and the previous discussion is still here, just as an 'outdated diff', which is sufficient, imo.

PR #104, opened 2012-10-05
Consistent error messages for API, rev2

jistr:
Validation errors follow the structure of other API errors. E.g.:

```
<error>
  <code>ValidationError</code>
  <message>Pool can't be blank;Name can't be blank</message>
</error>
```

Code for some specs and cucumber steps is updated accordingly.


jistr:
After a further discussion on IRC, this will not be merged in to prevent breaking compatibility with already existing APIs.  The error reporting consistency will be solved in a new version of API (after we implement API versioning).

PR #105, opened 2012-10-05
Introduces with_deleted method on paranoid objects

n1zyy:
It is functionally equivalent to calling "unscoped" (indeed, that's
all the method does), but it makes code using it far easier to
understand.


n1zyy:
Example usage: `Deployment.with_deleted.size` == `Deployment.unscoped.size`.

tzumainn:
Makes sense, ACK and merging

PR #109, opened 2012-10-08
Provider Types API - index action RM3880

petrblaho:
https://www.aeolusproject.org/redmine/issues/3880

Updates XML list of Provider Types via API - adds href attribute for each record.
Adds cucumber scenario for this action and rspec request tests.
Also removes cucumber feature file not needed anymore.


n1zyy:
ACK, merging to master.

PR #110, opened 2012-10-09
BZ 863383 - fixed migration of event description

jprovaznik:
rails 3.2 migration uses "def up" instead of "def self.up" which was used in older rails versions.
The problem is that "def up" is silently ignored in rails 3.0.x. Because of this
event an exception was raised when handling launch fail and an instance stayed in 'new' state.

https://bugzilla.redhat.com/show_bug.cgi?id=863383


humaton:
ACK.

PR #115, opened 2012-10-10
BZ864839 fix the wrong translation of pool title...

jtomasek:
...after backbone auto-update

https://bugzilla.redhat.com/show_bug.cgi?id=864839

### DIFF
--- a/src/app/controllers/pools_controller.rb
+++ b/src/app/controllers/pools_controller.rb
@@ -118,7 +118,7 @@ class PoolsController < ApplicationController
 
   def show
     @pool = Pool.find(params[:id])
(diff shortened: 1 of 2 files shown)


ifarkas:
Now the title changes from 'Default Pool' to 'Default' after 30 secs, so you might consider removing it from the pools.header_show.pool_name translation key too, just to make a consistent behaviour.

ifarkas:
Please also remove the pools.header_show.pool_name translation key because it's not used anywhere else.
Apart that, ACK

jtomasek:
The translation is used elsewhere (as we agreed), merging...

PR #117, opened 2012-10-10

rwsu:

Refactored https://github.com/aeolusproject/conductor/pull/71 to use
a provider_accounts resource under pool_families to manage the
associations.

Revised from https://github.com/aeolusproject/conductor/pull/100

### DIFF
--- a/aeolus-conductor.spec.in
+++ b/aeolus-conductor.spec.in
@@ -208,6 +208,7 @@ haml="app/views/hardware_profiles app/views/realm_mappings \
       app/views/provider_selections app/views/provider_priority_groups \
       app/views/roles app/views/providers app/views/settings \
       app/views/realms app/views/pool_families app/views/layouts\
+      app/views/pool_families_to_provider_accounts_associations\
       app/views/quotas app/views/permissions \
       app/views/deployments app/views/pools \
       app/views/instances app/views/user_sessions \
--- a/src/app/controllers/pool_families_to_provider_accounts_associations_controller.rb
+++ b/src/app/controllers/pool_families_to_provider_accounts_associations_controller.rb
@@ -1,0 +1,80 @@
+#
+#   Copyright 2012 Red Hat, Inc.
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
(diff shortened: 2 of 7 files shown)


jistr:
Strangest thing, GET on one specific association account returns 404 instead of 204: http://pastie.org/5034747

For me, the `...where(:id => provider_account.id).size` returns 2 on ProviderAccount with id 1. I have no idea why. When I print the whole array, it seems to have just one element, so I don't understand why #size returns 2. Here's some inspection: http://pastie.org/5034724

Are you able to reproduce this?

Changing the line in question from `...where(:id => provider_account.id).size == 1` to `...where(:id => provider_account.id).size > 0` would help of course, but I still have no idea why this happens in the first place. :-o

jistr:
I missed that ^^ during the first review because I probably tested GET on ProviderAccount with a different id and it would never occur to me that it could make a difference.

jistr:
Haha look at this

http://pastie.org/5034834

jistr:
Ok here it is:

```
conductor_development=# select * from pool_families_provider_accounts;
 provider_account_id | pool_family_id 
---------------------+----------------
                   1 |              1
                   1 |              1
                   2 |              1
(3 rows)
```

I probably messed up the DB myself manually when doing the last review. Your API does not create these double records, I tested for it right now.

ACK and merge :)

jistr:
Moving discussion from https://github.com/rwsu/conductor/commit/552e85cef20966e014c80333e5395971426f4c85 to here, as this place is probably more easy to locate in the future.

> Do what I think jstranks is suggesting and return a 200 when the
> server makes changes to the resource that is not explicitly set in the
> request.

No I wasn't suggesting that. The subresource has no XML representation, it returns 204 for GET as well. This is a very specific kind of subresource, it handles a plain has_and_belongs_to_many association. The only information contained in this subresource is its presence or absence, nothing more.

The discussion which lead up to this architecture is here: https://github.com/aeolusproject/conductor/pull/71 The approach is based on GitHub APIs.

> > return 204 when the association was already not present (code in pull rq already returns 204)
> 
> No.  This should return a 404.

Well... not sure. It would be cleaner wrt HTTP spec, but we would lose **idempotency** of the operation. What we're doing here is basically mimicking behavior of proposed LINK/UNLINK HTTP methods, which should be idempotent (http://tools.ietf.org/id/draft-snell-link-method-00.html). Since these methods are just draft and can't be used, we're using PUT/DELETE instead, but the semantics are more like LINK/UNLINK. So I'm on the fence here.

> Clients could PUT many associations at multiple levels in a since request.

I presume "since request" is a typo and you meant "single request". Currently, this API design allows creation/deletion of associations just one at a time. I thought that since it isn't an issue for GitHub, it wouldn't be an issue for us. Should we support multi-link and multi-unlink?

If you have on your mind something superior to this solution, please describe how should it behave (and if GET on a single instance of subresource should really be represented as XML, what would the XML look like), so that I can make a clearer picture :)

The important thing to note is that _we are not modelling provider accounts, we are modelling associations to provider accounts_.

E.g. in GitHub's case, you have

```
/users
/teams
/teams/id/members
```

where the `members` subresource is for creating/deleting associations (to `users`). There is no such thing as `/members` resource. I kinda like this modelling, but I'd be interested in other proposals :)

PR #119, opened 2012-10-11
unique index to pool_families_provider_accounts

martinpovolny:
add unique index to pool_families_provider_accounts to prevent model corruption as seen here: 
https://github.com/aeolusproject/conductor/pull/117


martinpovolny:
have to fix the tests

PR #122, opened 2012-10-15
unique index to pool_families_provider_accounts

martinpovolny:
.. and fix tests


jprovaznik:
ACK
(as we already discussed this is mostly test related bug, it shouldn't occur in other environments so there is no need to check existing data for duplicit records)

PR #125, opened 2012-10-15
Simplifying forms (round 1)

ifarkas:


jcoufal:
Checking just visual part.

##### Catalogs
- In heading, there is missing catalog icon. We should have it or not for all forms, but in other forms it is there, so it would be good to keep it here as well. In general icons reconstruction we can decide how to deal with them, but we should unify content in headings for forms now.
- Heading should stand for just "New Catalog" according to other headings.
- From Pool selection I would remove the first item "Select pool" and leave it blank, cause it is duplicating label (furthermore error message says that field can't be blank).

##### User group
- Do we support Membership source now? If we do, then ok. If not, and I don't see other option now (LDAP?), then remove whole row.
- For desciption would be better textarea.
- At edit form there is no icon in heading.

##### Pool families - Environments
- Use Environments instead of Pool Families, since we renamed that in app and button says "New environment".
- For max running instances use unlimited checkbox as we agreed - it is already styled in converge-ui. (-> remove note)

##### Pools
- For Max runn. inst the same as in Environments.
- Enabled is missing class at .control_group. There should be _.checkbox_ (it will fix the formatting).

##### General:
- After validation, when error occures, headings disappear.
- Some classes at control_group are extra and not supposed to be there (string, select, integer, ...)

##### Task for @jcoufal
- Fix spacing for select input In converge-ui.

jcoufal:
Update:
Spacing for select already fixed by last merge in converge-ui: https://github.com/Katello/converge-ui/pull/97 (hash: 365d5aa15862c16efa9ab7e134a4930a1eb2bb05)

ifarkas:
@jcoufal Thanks for the review. A few notes inline:

> ##### Catalogs
> - In heading, there is missing catalog icon. We should have it or not for all forms, but in other forms it is there, so it would be good to keep it here as well. In general icons reconstruction we can decide how to deal with them, but we should unify content in headings for forms now.

We don't have icons for all of our resources but I would rather not remove the existing icons. They improve the overall design of the page and they make a huge difference. How about rather requesting the missing icons?

> ##### User group
> - Do we support Membership source now? If we do, then ok. If not, and I don't see other option now (LDAP?), then remove whole row.

We do, you just have to configure it.

> - At edit form there is no icon in heading.
>   The same here as for the catalogs icon.
> 
> ##### Pool families - Environments
> - Use Environments instead of Pool Families, since we renamed that in app and button says "New environment".

As we discussed the renaming should happen in one step which _also_ takes care of renaming the model/controller/database table etc. Also worth noting that there is an ongoing debate about unifying the product and upstream names, so I would rather wait and not to change anything until the discussions are settled.

> - For max running instances use unlimited checkbox as we agreed - it is already styled in converge-ui. (-> remove note)

That discussion is only started _after_ I created the pull request. How I am supposed to use a design which were only created after the pull request and wasn't even considered before? Anyway, I will update the pull request, just wanted to point out the reason why it's not there already.

> ##### Pools
> - For Max runn. inst the same as in Environments.

The same here too.

> ##### General:
> - After validation, when error occures, headings disappear.

Could you be more specific here? I can't see any difference in the layout, the heading remains as it is, only the flash message is displayed along with the form showing the errors. Here's a screenshot of a form with validation errors: http://www.diigo.com/item/image/3aad2/uyog

> - Some classes at control_group are extra and not supposed to be there (string, select, integer, ...)

These are automatically generated by simple_form but as far as I know, we don't use those class, so they won't make any trouble for us. Or do they? Anyway, I will take a look at if we can easily remove them but if not, I would leave it as it is.

jcoufal:
> > ##### Catalogs
> > - In heading, there is missing catalog icon. We should have it or not for all forms, but in other forms it is there, so it would be good to keep it here as well. In general icons reconstruction we can decide how to deal with them, but we should unify content in headings for forms now.
> 
> We don't have icons for all of our resources but I would rather not remove the existing icons. They improve the overall design of the page and they make a huge difference. How about rather requesting the missing icons?

Yeah, I would leave them as well for now. Problem is that all icons are supposed to be redone (according to Westford talks) and nobody knows when. Are there lot of them missing? Sin thought - what about displaying "content icon" for catalogs? :)

> > ##### User groups
> > - At edit form there is no icon in heading.
> 
> The same here as for the catalogs icon.

We have icon for user group. In "new form" it is displayed, just missing for edit form.

> > ##### Pool families - Environments
> > - Use Environments instead of Pool Families, since we renamed that in app and button says "New environment".
> 
> As we discussed the renaming should happen in one step which _also_ takes care of renaming the model/controller/database table etc. Also worth noting that there is an ongoing debate about unifying the product and upstream names, so I would rather wait and not to change anything until the discussions are settled.

I know, but it is not logical for me, why we changed names to Environment in Admin > Environments, we are creating New Environment (according to button) and stating in label that we want Pool Family Name for it. This is just really crazy for me :(

> > - For max running instances use unlimited checkbox as we agreed - it is already styled in converge-ui. (-> remove note)
> 
> That discussion is only started _after_ I created the pull request. How I am supposed to use a design which were only created after the pull request and wasn't even considered before? Anyway, I will update the pull request, just wanted to point out the reason why it's not there already.

I know, I am aware of that. I didn't want to say that it is mistake, so if it sounds in that way, then I am really sorry. I just wanted to say, if you can take a look on that, please, and implement it in the forms as well, so these four are finished completely.

> > ##### General:
> > - After validation, when error occures, headings disappear.
> 
> Could you be more specific here? I can't see any difference in the layout, the heading remains as it is, only the flash message is displayed along with the form showing the errors. Here's a screenshot of a form with validation errors: http://www.diigo.com/item/image/3aad2/uyog

Sure, this problem appears at New/Edit Catalog, New/Edit Environment, New/Edit User Group, Edit Pool (New Pool works well). http://tmp.jaromircoufal.cz/missing_heading.png

Just curious, the screenshot you sent is from New User which is not part of this pull-request, right?

> > - Some classes at control_group are extra and not supposed to be there (string, select, integer, ...)
> 
> These are automatically generated by simple_form but as far as I know, we don't use those class, so they won't make any trouble for us. Or do they? Anyway, I will take a look at if we can easily remove them but if not, I would leave it as it is.

No they don't affect anything (what I found yet), they are just messing around there. I agree that if they are easy to remove it would be great, otherwise no problem.

ifarkas:
> > > - Some classes at control_group are extra and not supposed to be there (string, select, integer, ...)
> > 
> > These are automatically generated by simple_form but as far as I know, we don't use those class, so they won't make any trouble for us. Or do they? Anyway, I will take a look at if we can easily remove them but if not, I would leave it as it is.
> 
> No they don't affect anything (what I found yet), they are just messing around there. I agree that if they are easy to remove it would be great, otherwise no problem.

Unfortunately, I wasn't able to find an easy way to remove them. Apart that, all issues are fixed.

jcoufal:
Catalog and User Group are missing part of your update (as we discussed). For the rest:

New/Edit Pool:
- ~~Enabled (should contain class "checkbox" which will fix the horizontal alignment)~~ works

New/Edit Environment:
- For number type it might worth to set min="1" max="X" attribute in HTML tag, I don't see any hint message for it nor error message when I put negative number there.

ifarkas:
Sorry, somehow I messed up my patches. Now, it contains the first two patches too. I also improved the unlimited quota input as requested in the last comment.

jtomasek:
Works for me, ACK if @jcoufal has nothing to add. Just a tiny nit as we discussed: using button id instead of value in cucumber tests enables the test to run correctly with different language set, I can see that the tests are run in English and the button value is more readable so I'll leave it on your consideration.

jcoufal:
Just for Descriptions would be better to use "textarea" as I wrote in the very first comment. No blocker, can be fixed after applying this patch, but would be better to fix then. Everything looks that works, visual part is ok. ACK from my point of view, but wait for Jirka's more technical review.

EDIT: Jirka already reviewed, you can merge ;)

PR #127, opened 2012-10-16
Set the correct compass gem version constraints

jtomasek:


jprovaznik:
ACK this solved my issue

jguiditta:
I don't see why this is needed, all this information is already in the lock file (which this patch should have updated if there was an issue):

```
compass (0.11.7)
      chunky_png (~> 1.2)
      fssm (>= 0.2.7)
      sass (~> 3.1)
    compass-960-plugin (0.10.4)
      compass (>= 0.10.0)
```

jtomasek:
IMHO it is good to have a note in the Gemfile that if we update compass to 0.12 and later, we will get an issue. This way the possible bundle update in the future will keep the conductor in good shape or tells us that if we want newer version of compass, we should expect issues.

jguiditta:
That is valid reasoning, but otoh, he whole point of the lock file is to specify the exact version you know your app to work with.  An update of _any dependency_ should not be done without verifying the newer version is compatible with the application.

martinpovolny:
I understand this the same as dependencies in .deb or .rpm control files where you can also use both upper and lower limits or even declare conflicts.

If we know that some version of the package will break the app, then we can declare that in the Gemfile. Generally we would expect newer versions of gem to work with our app.

I see Gemfile.lock as a sort of guarantee. We are sure that the app will work with these versions, we have tested that.

At certain time someone will have to revisit the dependency and either provide a fix to the gem or to conductor that will make them compatible again.

Therefor I would accept the patch.

jistr:
This dependency constraint belongs into converge-ui's gemspec. Unfortunately, converge-ui doesn't have a gemspec :)

PR #130, opened 2012-10-17
BZ #860428: remove realm selection from deployment overview page

ifarkas:
https://bugzilla.redhat.com/show_bug.cgi?id=860428

### DIFF
--- a/src/app/views/deployments/_overview.html.haml
+++ b/src/app/views/deployments/_overview.html.haml
@@ -1,11 +1,13 @@
 - if @deployment.errors.any?
   = render 'layouts/error_messages', :object => @deployment
 = form_for @deployment, :html => {:class => 'generic horizontal'} do
+  = hidden_field :deployment, :name
   = hidden_field :deployment, :pool_id
   = hidden_field :deployment, :frontend_realm_id
   = hidden_field_tag :launch_parameters_encoded, @launch_parameters_encoded
   = hidden_field_tag :deployable_id, params[:deployable_id]
   = hidden_field_tag :backlink, @backlink
+
 
   %header.page-header
     %h1.pools= t('.pool_name', :pool => @pool.name)
@@ -24,7 +26,7 @@
         .field
           = label_tag :name, t('.name'), :class => 'em'
           .input
-            = text_field :deployment, :name, :class => 'em long'
+            %span.text_instead_input= @deployment.name
 
         - unless @deployment.deployable_xml.description.blank?
           .field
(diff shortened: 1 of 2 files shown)


jcoufal:
Partial ack. Process is OK now, just two small notes:
- Don't allow to change deployment name in Overview - display it.
- I named deployment "mock_deployment_2" and in Overview I got "mock-deployment-2" (with dashes), so we modified the name user chose. We should't do that.

ifarkas:
> - Don't allow to change deployment name in Overview - display it.
> - I named deployment "mock_deployment_2" and in Overview I got "mock-deployment-2" (with dashes), so we modified the name user chose. We should't do that.

@jcoufal Have you checked the linked BZ ticket? They specifically asked to remove the realm selection. While I agree that there's a lot needs to be changed regarding deployment launching but since this should also go to 1.1, I would keep this patch as compact and simple as it can be. 
I would prefer to have the deployment launch refactoring during some feature implementation and not with some bugfixing patch.

src/app/views/deployments/_overview.html.haml
```diff
@@ -24,7 +24,7 @@
         .field
           = label_tag :name, t('.name'), :class => 'em'
```
jcoufal, line 27: With change for .text_instead_input, the class for label should change to .field according to similar rows. (it is affecting padding, which is anyway wrong, but it should be at least same at all fields)

jcoufal:
@ifarkas Yeah, I did look and your previous patch fixed the BZ ticket. I just saw related small issue - that you are able to change name in overview as well, so I think that it should be corrected too. Anyway thank you for fixing that, even though it was not your task.

Just one little correction in above comment.

2 cucumber tests failed :(
Failing Scenarios:
cucumber features/deployment.feature:29 # Scenario: Launch new deployment
cucumber features/deployment.feature:49 # Scenario: Launch new deployment over X

ifarkas:
@jcoufal Ahh, I was mainly talking about the name conversion (underscore vs dashes) which I am not really aware if it is a feature or bug and what would be the effect of changing it, I would rather not risk to change it as it is also going to 1.1 branch.
Sorry for the confusion!

jtomasek suggested that instead of removing the realm selection, we could display it as a text, so I changed that. The same for the deployment name as you suggested. 
The failing tests are also fixed.

> With change for .text_instead_input, the class for label should change to .field according to similar rows. (it is affecting padding, which is anyway wrong, but it should be at least same at all fields)

Could you explain this a bit more? As I can see, in the other rows the labels don't have .field class but the label and the input field are wrapped in a div with a class .field, just as mine. What am I missing here?

sseago:
ACK this seems to do what the bug report calls for. It's a bit awkward if there's a matching error -- you get a "confirmation" page with a disabled 'launch' and nothing to do but go back or cancel, but it was this way before, too. Before it was worse, since you had a form with form elements and no way to submit them.

ifarkas:
@jcoufal, to address your concerns about the styling I created a new pull request: https://github.com/aeolusproject/conductor/pull/137

PR #131, opened 2012-10-17
BZ865548: fix autoupdating image list for given environment

ifarkas:
https://bugzilla.redhat.com/show_bug.cgi?id=865548


n1zyy:
It's too bad that these things require so much JavaScript to do the right thing, but this works perfectly. ACK.

PR #133, opened 2012-10-19
Fix rake dc:setup crash

jistr:
`rake dc:setup` used to crash with:

```
undefined method `find_by_username' for #<Class:0x000000059b3ac0>
```

The issue is that User db table is defined by migration to have `login`
attribute, and this attribute is renamed to `username` by a later
migration. However, this change is not reflected in `User` class.

This commit reloads model classes after all migrations are done, which
fixes the issue.

Fixes RM3903.


hacxman:
works for me, ack

PR #134, opened 2012-10-19
Sync Gemfile.lock with Gemfile

jistr:
There are version constraints for compass gem in Gemfile that should be
also in Gemfile.lock.

Right now, when I run `bundle`, Gemfile.lock gets regenerated differently than what's present in repo, which shouldn't happen.


jguiditta:
ACK

jtomasek:
Thanks for that

PR #135, opened 2012-10-19
Attempt to fix aeolus user's home directory even if logged in (RHBZ#868252)

jeckersb:


eggmaster:
So this works on f17, but I don't think the usermod is complaining there as we expect it to on RHEL, so I can't really test the functionality. I will try some other stuff on this but that's where I'm at with it right now.

PR #136, opened 2012-10-24
Provider Realm API

rwsu:
Implements index and show.

Also lists provider realms in provider and provider account show.


jistr:
Just one small thing - renaming the admin tab in UI introduced a small bug. If you go to Administer -> Cloud Providers -> pick one provider -> Provider Realms -> pick one realm. Then, clicking on "Return to Provider Realms" does not return you to provider realms, while on master it does. I would suggest making a separate pull request or maybe just a separate commit for the name change. Also, I believe the tab should be named `provider_realms`, not `frontend_realms`.

Otherwise this is good stuff, everything worked fine for me :) Conditional ACK with regard to ^^.

PR #137, opened 2012-10-24
Fix deployment overview styling

ifarkas:


jtomasek:
ACK, fixes the issue where the text_instead_input is slightly off in the form

PR #146, opened 2012-10-26
Password and Username recovery

jtomasek:
This patchset adds Username and Password recovery functionality to conductor.

To make emails work properly, it is necessary to update configuration in src/config/settings.yml (for developement) or /etc/aeolus-conductor/settings.yml (for production). 

It is necessary to set the action_mailer.default_from to valid and existing email address, to make the email actually sent. Next action_mailer.default_url_options.host and protocol serve for the links used in email texts. If one of these is set wrong, the links used in emails (like link to page for changing the password using password reset token) will not lead to the right url.

### DIFF
--- a/aeolus-conductor.spec.in
+++ b/aeolus-conductor.spec.in
@@ -225,7 +225,8 @@ haml="app/views/hardware_profiles app/views/realm_mappings \
       app/views/pool_families_to_provider_accounts_associations\
       app/views/quotas app/views/permissions \
       app/views/deployments app/views/pools \
-      app/views/instances app/views/user_sessions \
+      app/views/instances app/views/user_sessions app/views/password_resets \
+      app/views/user_mailer
       app/views/deployables app/views/catalogs \
       app/views/provider_types app/views/api/images \
       app/views/config_servers app/views/provider_realms \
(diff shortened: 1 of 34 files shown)


jtomasek:
Note that with this patchset applied, you will experience fonts not loaded properly. It is because of change in converge-ui (symlink has to be changed), I will send patch to fix this issue after this is pushed in.

jtomasek:
@n1zyy Thanks for the notes. I have updated the patchset accordingly.

Regarding the success notification after username and password recovery, I have added descriptive comment to the code that we are doing this on purpose (safety reasons). I know this solution is not perfect. I am planning to add at least some validations (validate that all fields have been filled) to these forms in subsequent patches for this feature.

> Is this config needed here, where delivery_method is :test?

Yes it is. The default_url_options have to be set in test env. too.

> I wonder if we should ask aeolus-configure to help automate this. Though I suppose the best we could do would be to default to the FQDN, and default_from below would be a tough default.

I have discussed ways how could aeolus-configure help with automating this configuration with eck and jprovazn. I would like to focus on this after this patchset is in.

> I think this file might be something that users now need to be instructed to configure?

Yes, for now to make emails work it is necessary to configure it in the settings.yml properly. (as described in intro).

n1zyy:
Thanks! This looks fantastic now. I was apparently having some local MTA problems as well, but your SMTP config helped me work around that.

ACK, will merge.

Two things to note on the commit:
- The converge-ui update causes Overpass to stop working. (You noted this.)
- The converge-ui update also causes the header to become a lighter shade of blue. (This is because it's changed in converge-ui, and should be fixed separately, not as part of your patch.)

PR #150, opened 2012-10-26
Make dc:setup print admin user password  to console

mfojtik:
Last night I was trying to setup Conductor from GIT on my machine and I forgot the default 'admin' password (created by "rake dc:setup"). This patch just simply print the password to console, so new contributors will not need to look into code or search on site for it :-)


mfojtik:
ping?

PR #151, opened 2012-10-27
Provider Realm API

rwsu:
Implements index and show.

Also lists provider realms in provider and provider account show.

Fixes issue found in a previous pull request, https://github.com/aeolusproject/conductor/pull/136.

### DIFF
--- a/src/app/controllers/provider_realms_controller.rb
+++ b/src/app/controllers/provider_realms_controller.rb
@@ -21,26 +21,27 @@ class ProviderRealmsController < ApplicationController
   def index
     clear_breadcrumbs
     save_breadcrumb(provider_realms_path)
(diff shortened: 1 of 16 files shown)


jistr:
ACK, works for me :) I'm not merging it in yet, in case you want to clean away the changes in commit e9686668 (those that get reverted by the later commits). Basically this would be done by squashing fbfe8a9 into e9686668 and removing the set_admin_content_tabs change from e9686668. No need to submit it as pull rq again, just push it :)

rwsu:
Thanks for the reviews. I folded the last two changes into the first and pushed the first two changes into master. I left out the third part which added provider realms to provider and provider account resources. Since we're looking at supporting puts of full resources, the update actions will need further enhancement that this patch didn't address.

PR #152, opened 2012-10-28
add size limit to image template new from upload -- issue #142

### LINKED ISSUE #142: image import should handle large input with a sane error message
## What I want

Talking about '/images/new?environment=1' -- "New Image": 

If I upload an ISO image, which is what some user will surely try, I want the application to give me a nice error message.

Generally when I give the application a wrong input (wrong type, wrong size) I expect a nice error message.
## What I have

After a dozen minutes I have an empty browser window. And in the Rails log I see:

```
              Rendered layouts/_admin_nav.html.haml (19.5ms)
              Rendered images/edit_xml.html.haml within layouts/application (365004.5ms)
              Rendered layouts/_new_notification.html.haml (45.3ms)
              Rendered layouts/_footer.html.haml (882.3ms)
              Rendered layouts/converge-ui/_header_layout.haml (4.9ms)
              Rendered layouts/converge-ui/_base.haml (41565.7ms)
              Rendered layouts/converge-ui/shell_layout.haml (201594.5ms)
              Completed 200 OK in 753340ms (Views: 743206.8ms | ActiveRecord: 5.9ms)
              terminate called after throwing an instance of 'std::runtime_error'
              what():  no allocation for outbound data
              Aborted
```

The UI has changed since my last week experiments so that I no longer have a strong urge to upload an ISO ;-)

But anyway the application should handle that case.

(rand issue # 6)

martinpovolny:
add size limit to image template new from upload -- issue #142

closes #142

### DIFF
--- a/src/app/controllers/images_controller.rb
+++ b/src/app/controllers/images_controller.rb
@@ -16,6 +16,7 @@
 
 class ImagesController < ApplicationController
   before_filter :require_user
+  before_filter :check_request_size, :only=> :edit_xml
 
   def index
     set_admin_environments_tabs 'images'
@@ -369,6 +370,12 @@ class ImagesController < ApplicationController
   end
 
   protected
+  def check_request_size
+    if request.headers["CONTENT_LENGTH"].to_i > 31457280
+      redirect_to request.referrer, :flash => { :error => t('images.flash.error.too_large') }
+    end
+  end
+
   def load_target_images(build)
     @target_images_by_target = {}
     return unless build and @latest_build_id.present?
--- a/src/config/locales/en.yml
+++ b/src/config/locales/en.yml
@@ -1000,6 +1000,7 @@ en:
         delete_failed: Unable to Delete Image
         not_found: Image not found
       error:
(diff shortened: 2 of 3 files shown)


jprovaznik:
With this patch, data are still uploaded to server, but at least it doesn't try to load it as XML document. So this patch is partial solution, but better than nothing - I'll ACK it once you attach a test for this patch (unless it's too complicated to write such test).

For better fix, we would need request size checking on rack (or apache) level and also on client side, client side checking should be already be possible in newer browsers: http://www.html5rocks.com/en/tutorials/file/dndfiles/

martinpovolny:
Yes, it's a solution for #142 -- prevents the server from crashing. 

It's not a general solution for preventing upload of data of insane size.

That type of check would have to be done on lower level, probably as a part of deployment process. 

RPM or other installs can solve this as part of configuring Passenger or whatever Rails will be running on.

I will add the test, as you request.

jprovaznik:
ACK, will cherry-pick also into tim branch to make sure this will not be removed

Issue #149, opened 2012-10-26
inconsistent terminology in the New Environment page


martinpovolny:
at /pool_families/new I see section title: "New Environment"

but below a label for a text field reads: "Pool Family Name"

razorinc:
And you can't send a patch for it?

aathomas:
This is a tricky one. 

Pool Families is definitely wrong, and we need to rename the controllers etc..

However, there are problems with "Environment" too (See the recent thread on the developer list about reconciling names)

So, it isn't clear what the correct name should be. It will become clear soon, though, and, once we have a new name, it should be applied consistently, thoughout the code, the APIs, and the UI.

jcoufal:
I think that @razorinc wanted to point out, that sometimes it's faster to fix issue right away, than filling tickets.

Regarding renaming all components, it's more tricky as @aathomas mentioned.

But right now in the UI there are "Environments" everywhere so it is obvious that Pool Family should be changed. Until we get into real renaming (will take some time), this quick fix in the UI would be useful.

aathomas:
@jcoufal is correct. For the time being, the correct name to use consistently in the UI is "Environments"

martinpovolny:
@jcoufal: yes, sometimes it's faster, sometimes not

for me atm, it's not

I am trying to post issues that I come across while doing other work and even reporting the issue breaks me from the thing I work on

but enough of this

razorinc:
@jcoufal That's exactly what I meant.
@martinpovolny Yeah, I understand but delegating does not actually always work, I think would be better if you can write a quick patch and see how it's going.

PR #153, opened 2012-10-28
fix issue #149 - label in pool_families/new form 

martinpovolny:
fix issue #149 - label in pool_families/new form

### DIFF
--- a/src/app/views/pool_families/_form.html.haml
+++ b/src/app/views/pool_families/_form.html.haml
@@ -2,7 +2,7 @@
   = render 'layouts/error_messages', :object => @pool_family
 %fieldset
   .field
-    = form.label :name, :class => 'em'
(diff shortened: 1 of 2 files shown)


humaton:
ack.

jtomasek:
NACK, the label in the form is label for actual activerecord model attribute. So all you have to do to get proper label is to fix the translation in config/locales/activerecord/en.yml (line 223)

humaton:
Ahh my fault I wasn't myself today morning. How do I revert a merge here?

jtomasek:
I am not sure about github feature, but either create git revert patch or better solution I think is just to create patch that fixes this one and introduces the correct change... I can do that, if you want.

humaton:
That will be awesome, thanks.

jtomasek:
fix: https://github.com/aeolusproject/conductor/pull/154

PR #154, opened 2012-10-29
This path fixes 8fd0330 and introduces correct translation

jtomasek:
see https://github.com/aeolusproject/conductor/pull/153


humaton:
ACK Finally fixed \o/

PR #159, opened 2012-11-01
remove unused model: InstanceHwp

martinpovolny:
the model InstanceHwp is not used in the code

this patch removes the unused model, drops the table and drops the reference from Instance


martinpovolny:
instead of removing the model I follow up with a patch to actually fix the model and start using it

PR #162, opened 2012-11-01
Converge-ui fonts symlink fix

jtomasek:
Latest converge-ui returns back to original fonts folder structure. So the converge-ui
fonts are now symlinked in public/fonts/converge-ui. If we have custom fonts in
Conductor, we should put them into public/fonts dir.

### DIFF
--- a/aeolus-conductor.spec.in
+++ b/aeolus-conductor.spec.in
@@ -271,7 +271,10 @@ rb="app/models app/controllers app/controllers/api app/helpers app/mailers app/s
     vendor/provider_selection/strategies/strict_order \
     vendor/provider_selection/strategies/penalty_for_failure"
 rhtml="app/views/layouts"
-svg="public/images public/images/icons"
+svg="public/fonts public/images public/images/icons"
+ttf="public/fonts"
+eot="public/fonts"
+woff="public/fonts"
 txt="public"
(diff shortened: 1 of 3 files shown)


ifarkas:
The rpm building throws the following error: 
app/stylesheets/layout.scss (Line 4: File to import not found or unreadable: converge-ui/composites/content_elements.

Not sure if it's related but it seems that the converge-ui update is causing the failure.

jeckersb:
This builds ok for me, but I'm worried that it's going to break because the fonts end up being nested a directory deeper than they were before.

Before:

/usr/share/aeolus-conductor/public/fonts/LiberationSans-Bold-webfont.eot -> ../../vendor/converge-ui/fonts/LiberationSans-Bold-webfont.eot

After, only the directory is linked:

/usr/share/aeolus-conductor/public/fonts/converge-ui -> ../../vendor/converge-ui/fonts

Which means the above font is now canonically at /usr/share/aeolus-conductor/public/fonts/converge-ui/LiberationSans-Bold-webfont.eot (note the new addition of /converge-ui/ in the path)

jtomasek:
Note (regarding IRC discussion): It will not break because the links in converge-ui changed, so the symlink this patch introduces is actually making the fonts work with the new converge-ui

ifarkas:
I needed to build and install the converge-ui rpm from source and it solved the issue.
ACK

PR #166, opened 2012-11-12
make taskomatic create instances of InstanceHwp

martinpovolny:
- fix taskomatic so that it creates instances of InstanceHwp
- add necessary link to hardware_profile to the model


hacxman:
ack

PR #161, opened 2012-11-01
API consistency - resources linking, provider accounts creation

jistr:
Another progress on API consistency. See commit messages for details.

All changes are meant to be backwards compatible.

PR #169, opened 2012-11-14
API consistency - resources linking, provider accounts creation v2

jistr:
Pull request #161 with added licensing and docs.


jistr:
Conditional ACK for this was issued in pull request #161, build passed, merging.

PR #172, opened 2012-11-16
Issue 171 - Instances api - list, show, entrypoint

rwsu:
A few issues to be addressed in a future patch:
1. Assembly xml is shown as quoted.
2. Image uuids should be replaced with references to actual resources.

### DIFF
--- a/src/app/controllers/instances_controller.rb
+++ b/src/app/controllers/instances_controller.rb
@@ -30,6 +30,7 @@ class InstancesController < ApplicationController
       format.html
       format.js { render :partial => 'list' }
       format.json { render :json => @instances.map{ |instance| view_context.instance_for_mustache(instance) } }
+      format.xml { render :partial => 'list.xml' }
     end
   end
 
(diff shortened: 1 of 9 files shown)


hacxman:
Hi, some of the lines are too long, please break it down so that we could see it without wrapping.
In view it's sufficient to keep the indentation level.

Also it might be worth to split _it_ in spec so that each _should_ is in it's own _it_. It's not a blocker though.

Other than that it looks OK to me, so conditional ACK.
Have a nice day.

hacxman:
Well, unfortunately I found out that it's not worth to split _it_ into a lot of new _it_s because of speed.
When I split _it_s when testing XML content, the spec took 3.5 times longer :(

rwsu:
This patch needs to be revised to show the entire resource representation in top level collections. I am planning to submit another pull for the next milestone.

rwsu:
Closing. Replaced by pull #273.

PR #175, opened 2012-11-20
add hardware profile to matches for provider selection

martinpovolny:
on the road to enable cost-based provider selection this small patch adds hardware profile to each match for provider selection so that provider selection strategies can look not only at provider accounts but also at the hardware profiles when calculating the scores

please do not leave this pull request hanging like the 2 trivial pull request that I have created 3 weeks ago

### DIFF
--- a/src/lib/provider_selection/base.rb
+++ b/src/lib/provider_selection/base.rb
@@ -55,26 +55,25 @@ module ProviderSelection
       pool = @instances.first.pool
       rank = Rank.new(pool)
 
-      provider_accounts = find_common_provider_accounts.map do |provider_account|
-        provider_account if provider_account.quota.can_start?(@instances)
-      end.uniq.compact
-
       # Adding a priority of 100000 to the default priority group which contains
       # all the available provider accounts.
       # The user defined priority groups has a score between -100 and +100.
       default_priority_group = PriorityGroup.new(100000)
       rank.default_priority_group = default_priority_group
 
-      provider_accounts.each do |provider_account|
+      find_common_provider_accounts.each do |acc_with_hwp|
+        next unless acc_with_hwp.provider_account.quota.can_start?(@instances)
+
         # Rescale to provider account priority to the [-100, 100] interval
-        score = provider_account.priority
+        score = acc_with_hwp.provider_account.priority
         if score.present? && score > Match::UPPER_LIMIT
           score = Match::UPPER_LIMIT
         elsif score.present? && score < Match::LOWER_LIMIT
           score = Match::LOWER_LIMIT
         end
 
-        default_priority_group.matches << Match.new(:provider_account => provider_account,
+        default_priority_group.matches << Match.new(:provider_account => acc_with_hwp.provider_account,
+                                                    :hardware_profile => acc_with_hwp.hardware_profile,
                                                     :score => score)
       end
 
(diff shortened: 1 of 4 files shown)


src/lib/provider_selection/base.rb
```diff
@@ -109,21 +108,30 @@ def chain_strategy(name, options = {})
...
 
+    ProviderWithProfile = Struct.new(:provider_account, :hardware_profile)
+
+    class SetOfProviderWithProfile < Array
+      def intersect!(set)
+        provider_accounts = set.map(&:provider_account)
+        self.delete_if{ |el| ! provider_accounts.include?(el.provider_account) }
+      end
+    end
+
```
ifarkas, line 119: Defining ProviderWithProfile struct and SetOfProviderWithProfile class after the private keyword is misleading. It has nothing to do with private because they are globally accessible names and should not pollute the ProviderSelection::Base class. Please move them to a separate file/files.

src/lib/provider_selection/base.rb
```diff
@@ -109,21 +108,30 @@ def chain_strategy(name, options = {})
...
       instance_matches_grouped_by_instances = @instances.map do |instance|
         filter_instance_matches(instance)
       end
 
-      result = []
+      profiles = SetOfProviderWithProfile.new
       instance_matches_grouped_by_instances.each_with_index do |instance_matches, index|
         if index == 0
-          result += instance_matches.map(&:provider_account)
+          instance_matches.map { |im| profiles << ProviderWithProfile.new(im.provider_account, im.hardware_profile) }
```
ifarkas, line 128: If you moved the intersect method to a different class, you may also consider moving the fill up code to its own method for readability.

src/lib/provider_selection/priority_group.rb
```diff
@@ -30,17 +30,9 @@ def self.create_from_active_record(obj, allowed_matches)
...
-      possible_provider_accounts &= allowed_provider_accounts
-
-      return nil if possible_provider_accounts.empty?
-
-      possible_provider_accounts.each do |provider_account|
-        priority_group.matches << Match.new(:provider_account => provider_account)
-      end
+      priority_group.matches = allowed_matches.find_all { |match| 
+        possible_provider_accounts.include?( match.provider_account )
+      }
```
ifarkas, line 35: The problem with your implementation is that it copies the references of the matches from the default priority group, instead of creating new match objects. As a side effect, if you change the score of a match in the default priority group, it would change the score in the other priority group too. This is not the behavior what we want.

ifarkas:
In general:
You should read the Coding Guidelines at https://aeolusproject.org/redmine/projects/aeolus/wiki/Coding_Guidelines and reformat your patch accordingly.

The naming you used is somewhat misfortunate. In a lot of places you are using the word 'provider' for 'provider account' but in Conductor we have both models. This may bring confusion to anyone who expects a provider because of its name and will receive a provider account. You may also consider using the same name for the same thing consistently: instead of referring to 'provider account' as 'provider' in certain cases but as 'acc' in other cases, you should stick to one. I would vote for using the whole name like 'provider account' just for clarity but it's up to you.

src/lib/provider_selection/base.rb
martinpovolny, line 119: The goal of the placement of the code inside the class is to provide a hint for someone that comes after me, that this is an internal class that hides an implementation detail. Therefor it is inside the other class, behind the 'private' keyword and not in a separate file.

I don't care much and not really insist on this, but creating too many files with too many simple classes does not help to read the code.

So if you insist, I will do as you say, but I do not agree.

src/lib/provider_selection/base.rb
martinpovolny, line 128: Sure I could. But I just swapped the array for the utility class I created. I did not want to refactor your code.

src/lib/provider_selection/priority_group.rb
martinpovolny, line 35: Ok, I'll fix this. Obviously I did not understand the original code to get this.

But no problem with the omitted 'uniq'? Not here and there was one more place where I dumped it. Was not completely sure...

martinpovolny:
Sure, you are right, I will fix the names.

Given I have red the guidelines, I have to ask what other violations of the guidelines did you find?

ifarkas:
The issues were discussed on IRC.

Issue #177, opened 2012-11-20
provider selection configuration UI broken


martinpovolny:
if I go:

Administer-->Environments-->POOL NAME-->Provider selection

then turn on "Penalty for Failure" and then click "Configure"

I get:
# actual result

uninitialized constant ApplicationHelper::FormBuilderWithRequiredFields
# expected result

UI for setting strategy specific options opens

PR #178, opened 2012-11-20
test for issue #177

martinpovolny:
after applying this patch, 'rake cucumber' fails

it should fail, because there is a bug -- issue #177

this patch tests previously untested path in the conductor ui, not only the page with issue #177


martinpovolny:
included as part  of #183

PR #182, opened 2012-11-22
add hardware profile to matches for provider selection

martinpovolny:
Imre, please check this variant

there's a potential difference in the behavior in the create_from_active_record method

there's a case where the original implementation would return PriorityGroup with empty matches and the new would return null

not sure if that is right

please check

### DIFF
--- a/src/lib/provider_selection/base.rb
+++ b/src/lib/provider_selection/base.rb
@@ -55,26 +55,25 @@ module ProviderSelection
       pool = @instances.first.pool
       rank = Rank.new(pool)
 
-      provider_accounts = find_common_provider_accounts.map do |provider_account|
-        provider_account if provider_account.quota.can_start?(@instances)
-      end.uniq.compact
-
       # Adding a priority of 100000 to the default priority group which contains
       # all the available provider accounts.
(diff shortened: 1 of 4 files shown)


src/lib/provider_selection/base.rb
```diff
@@ -107,23 +106,30 @@ def chain_strategy(name, options = {})
...
+          ProviderAccWithHwp.new(instance_match.provider_account, instance_match.hardware_profile)
+        end
+
         if index == 0
-          result += instance_matches.map(&:provider_account)
+          common_accounts = accounts
         else
-          result &= instance_matches.map(&:provider_account)
+          provider_accounts = accounts.map(&:provider_account)
+          common_accounts.delete_if{ |pair| ! provider_accounts.include?(pair.provider_account) }
```
ifarkas: Please remove the space after the negation operator.

ifarkas:
Returning nil from ProviderSelection::PriorityGroup.create_from_active_record won't cause any issue.

The tests are also passing for me, Travis is reporting incorrectly.

ACK with 1 inline note.

PR #184, opened 2012-11-23
Deployables API index

hacxman:


src/spec/requests/api/deployables_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    shared_examples_for "should return list of deployables" do
+      let(:xml) { Nokogiri::XML(response.body) }
+      let(:deployables) { Deployable.all } # assuming there are no other
+
+      it { xml.xpath('/deployables').size.should == 1 }
+      it { xml.xpath('/deployables/deployable').size.should == deployables.length }
+      it "with correct IDs" do
+        _ids = Deployable.all.map(&:id).map(&:to_s).to_set
+        xml.xpath('/deployables/deployable/@id')
+                 .map(&:value).to_set.should == _ids
```
rwsu, line 27: The specs failed to run on ruby 1.8.7. See travis build output.

I think the issue is the period should be moved up to the previous line. The line break would occur between '.' and 'm' instead of ')' and '.'.

src/spec/requests/api/deployables_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+      it "with correct hrefs" do
+        _hrefs = Deployable.all.map {|_d| deployable_url(_d)}.to_set
+        xml.xpath('/deployables/deployable/@href')
+                 .map(&:value).to_set.should == _hrefs
```
rwsu, line 32: Same issue here.

src/spec/requests/api/deployables_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      context "a catalog with some deployables" do
+        before do
+          _cid = FactoryGirl.create(:catalog_with_deployable).id
+          get "catalogs/#{_cid}/deployables", nil, headers
+        end
+
+        it_behaves_like "response should be success & return xml"
+        
+        it_behaves_like "should return list of deployables"
```
rwsu, line 66: We need to assert that the deployables that are listed are actually the ones for the catalog. Right now I think it assumes it returns Deployables.all.
We should add some more test data that includes deployables not associated with the catalog.

src/spec/requests/api/deployables_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      context "deployables associated with some pool" do
+        before do
+          _pool = FactoryGirl.create(:pool_with_catalog_with_deployable)
+          @depls = _pool.catalogs.map { |_c| _c.deployables }.reduce(&:+)
+          get "pools/#{_pool.id}/deployables", nil, headers
+        end
+
+        it_behaves_like "response should be success & return xml"
+
+        it_behaves_like "should return list of deployables"
```
rwsu, line 78: Same here. We should check that we are only showing deployables that are associated with the selected pool. Add a deployable not associated with the pool in the test.

rwsu:
I found a few issues. 

Also, the index action needs to be updated to return the full representation for top-level collections. This was decided upon in the previous tech cabal meeting. See https://redmine.aeolusproject.org/redmine/projects/aeolus/wiki/ConductorAPIResourceRepresentation. 

To conform to the new api format, I'll leave it to you to see if you want to incorporate the changes in an update of this pull request or in another pull.

hacxman:
Index action will be updated to return full representation _after_ show action is merged. It doesn't make sense to implement it sooner and depend on nonexisting feature.

hacxman:
Revised version available @ #206

PR #186, opened 2012-11-23
Deployments API read actions

jistr:
This implements read API actions for deployments: listings and show.

Two notes:
- Listing was implemented before Tech Cabal decision that top level listings should be deep and other listings shallow. This double-default nature of #index action will require both code and spec changes and I don't want to delay this pull request, so I'll implement it subsequently.
- # show action provides some basic information. Parts of #show action that require reading CIMI spec or implementation of some other resources are left out for now as well.

### DIFF
--- a/src/app/controllers/deployments_controller.rb
+++ b/src/app/controllers/deployments_controller.rb
@@ -242,6 +242,7 @@ class DeploymentsController < ApplicationController
       format.html { render :action => 'show'}
       format.js   { render :partial => @details_tab[:view] }
       format.json { render :json => @deployment }
+      format.xml
     end
   end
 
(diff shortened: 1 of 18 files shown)


rwsu:
Ack.

I liked that you removed pagination from the api for now. We'll need to revisit pagination in the other api resources.

jistr:
Thanks for the review, I pushed the commits :)

I hope I can finish some tweaks today and submit another pull request that would make deployments API behave according to the wiki page that Martyn created yesterday [1].

[1] https://redmine.aeolusproject.org/redmine/projects/aeolus/wiki/ConductorAPIResourceRepresentation

PR #199, opened 2012-11-26
Provider Types API - delete action

petrblaho:
https://redmine.aeolusproject.org/redmine/issues/3884

### DIFF
--- a/src/app/controllers/provider_types_controller.rb
+++ b/src/app/controllers/provider_types_controller.rb
@@ -30,4 +30,17 @@ class ProviderTypesController < ApplicationController
       format.xml
     end
   end
+
+  def destroy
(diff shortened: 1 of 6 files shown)


jistr:
Non-issues, just unimportant notes:
- Creating a non-persisted ProviderType instance when testing for deleting a non-existent provider type is faster if you use `Factory.build` and pass in `:id => -1`.
- I think it's common coding style not to put spaces on the inner side of parentheses, like `( bad + bad )` -> `(good + good)`.

Issues:
- Shouldn't there be some kind of permission check when deleting a provider type?

petrblaho:
I made a revision here - https://github.com/aeolusproject/conductor/pull/265

PR #200, opened 2012-11-27
costs for conductor

### LINKED ISSUE #192: Cost-based provider selection
UI for adding costs
UI for managing selection algorithm 
actual algorithm

martinpovolny:
implements:
- assign cost to backend hardware profiles
- calculate costs for instances, deployments etc
- simple cost-bases provider selection strategy

closes #192

### DIFF
--- a/src/app/controllers/hardware_profiles_controller.rb
+++ b/src/app/controllers/hardware_profiles_controller.rb
@@ -19,6 +19,7 @@ class HardwareProfilesController < ApplicationController
   before_filter :load_hardware_profiles, :only => [:index, :show]
   before_filter :setup_new_hardware_profile, :only => [:new]
   before_filter :setup_hardware_profile, :only => [:new, :create, :matching_provider_hardware_profiles, :edit, :update]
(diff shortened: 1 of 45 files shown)


jprovaznik:
If I set a cost value (no matter what billing model is selected), then final cost on hardware profile show page is always cost*2.

jprovaznik:
doesn't have to be part of this patch, but we will need to give users more info what range of cost values is expected - now I was wondering why I'm getting maximum penalty for two different costs (both in range 1-10), after checking the code it seems that expected range is 0.01-0.9

martinpovolny:
there is a constant in the code, that is meant to handle this -- if you set it to a higher value, you move the expected price range

there's a note that this might become a config option for the strategy if needed

so if there's a need (the prices actually would be needed way out of the expected range) I can turn that constant into a config option with minimal effort

jprovaznik:
re: cost values range - we discussed this over irc and agreed that information about what is reasonable cost value range should be displayed when editting costs. This will be fixed in some way in seprate patch.

PR #201, opened 2012-11-27
Simplifying provider_accounts form

ifarkas:


ifarkas:
Included in https://github.com/aeolusproject/conductor/pull/223

PR #206, opened 2012-11-29
Deployables index, second attempt

hacxman:
Index doesn't return full representation in the top level collection since show action is not yet implemented. It (full representation) will be added once the show action is merged.


rwsu:
This looks better. But I've spotted a couple more issues.
1. http://localhost/conductor/api/deployables route is missing. You're assuming http://localhost/conductor/deployables for api which is not correct.
2. This command returns zero deployables when some deployables are in the system.
   curl -X GET --user admin:password --header "Accept: application/xml" -k  https://localhost/conductor/deployables 

The subcollections under https://localhost/conductor/catalogs/1/deployables and https://localhost/conductor/deployments/1/deployables do appear to work. We'll also need to verify those under the api namespace are also working once you add that route.

hacxman:
new PR -> https://github.com/aeolusproject/conductor/pull/228

PR #223, opened 2012-12-10
Simplifying forms (round 2)

ifarkas:


jcoufal:
In general looks good, just few comments:

**Provider account:**
- ~~Restrict number of rows in textarea (2 rows? - depends on usual length of the key, but 2 should be enough).~~ migrate fixed that
- after submitting the empty form "key" and "certificate" disappears

**Provider:**
- ~~Cancel should get me at provider detail (unfortunately the same page which has the form) not to list of providers.~~ agreed we will fix this with following provider's page fix

**Frontend realms:**
- Description as textarea having 2 rows.
- At form remove "new realm" and "cancel editing" from form header
- (maybe consider to upgrade "Add mapping" forms as well in this section)

jcoufal:
(+ careful about failed Travis)

ifarkas:
@jcoufal, thanks for the review! All issues should be fixed by now. I also upgraded the add mappings form as you suggested.

jcoufal:
From the UI point of view, it looks good. Partial ACK - wait for some back-end check ;)

src/config/settings.yml
```diff
@@ -21,7 +21,8 @@
   #   :consumer_secret: consumer_secret_here
 
 :session:
-  :timeout: 15 # minutes
+  :timeout: 150 # minutes
+#  :timeout: 150000000000 # minutes
```
jtomasek: I assume this should be set back to 15 mins.

jtomasek:
ACK, great work! (tiny nit above)

ifarkas:
Fixed and merged. Thanks for the review for both of you!

PR #224, opened 2012-12-10
request versions of gems that work on ubuntu/debian

martinpovolny:
This pull request changes required versions of gems in Gemfile.lock so that the particular gems are possible to install on Debian and Ubuntu systems.


jguiditta:
I am not opposed to updating deps to newer versions, but the distro is not a driver for this.  If aeolus/aconductor is packaged for a given distro, it should be using Gemfile.in/bundler_ext, which means it uses system versions of gems.  If debian happens to have newer available and they work with our code, great.  This is not a nack, I just wanted to clarify when Gemfile.lock updates should be needed.

martinpovolny:
I am not talking about packaging for Debian and Ubuntu at this point. It's just about checking the sources out of the repo (including the Gemfile.lock), doing 'bundler install' and running Conductor on these distributions. I am not talking about using system versions of these gems.

The packages involved in versions that we use do not compile on Debian and Ubuntu due to more strict compiler settings. The packages are fixed in the versions that I request in such a way, that they actually compile with these stricter compiler settings.

While I don't understand why supporting more distros (and more users) is NOT a driver for updating gems (Is that what you are saying? Please correct me if not.) I hope that having the package improved (it compiles with more strict compiler settings) might be beneficial and may provide further reason for upgrading the package.

Please let me know if we understand each other.

jguiditta:
My point was that if , say, conductor were packaged for debian, it would use the newer thin that debian needs, for example, because that it what would be pulled in via debs and conductor would just load it.  This is irrelevant to the lock file version.  Version upgrades in the lock file should be driven by new features and/or bug fixes that we want to leverage in our code.  What a distro ships with is an orthogonal issue, and outside the scope/concern of the upstream project.  This is the entire point of the Gemfile/Gemfile.in setup.  Any installation should have  one or the other, but not both.  

Now, if you are doing a developer install on debian, and hitting issues compiling a gem in userspace vs system (which is perhaps the issue you are hitting?), that is a slightly different can of worms.  I have yet to hit such an issue, as gem requirements for c library versions (like libxml2, for example) usually are not that strict, so I am not sure how you would hit such a failure.  My suspicion is that you are doing some combination of a developer setup, and using debs, but perhaps I am misunderstanding.  To attempt to clarify, the 2 options for a setup should basically be:

1) Pure ruby developer setup: This setup is most likely to use rvm or rbenv (or even bundler with the --path option set to a local directory), and will use the git-controller Gemfile/Gemfile.lock

2) 'Platform developer' setup: This person prefers to use the gems/versions packaged for their chosen distribution.  In this case, they would (for development) mv Gemfile Gemfile.in, and just pick up the system-installed gems.  Typically the dependencies will be satisfied by doing an installation of aeolus-all via the package manger on that system.  any test dependencies a developer may add in those cases would either be added locally to their package manifest, installed directly via their package manager, or installed with the correct version via gem to the system installation location (wherever that may be on a given system).

If you are trying to cover some third case not listed above, perhaps that is the disconnect here, ad you could explain to me what that case is?

Lastly, I am again, not against a version update if there is benefit to the upstream codebase, but this should not be a decision made by native package availability in a given distribution.

martinpovolny:
Hi. A am in the situation 1) pure ruby dev. setup, you call it.

The problem is that the 3 ruby extensions/gems involved where written in a sort of "relaxed" way ignoring certain issues -- see here e.g. (https://github.com/eventmachine/eventmachine/issues/345#issuecomment-10869240). 

Because there is a system-wide default config for GCC that is more strict on Ubuntu and Debian the e.g. on Fedora this causes a problem that is triggered if I use system ruby (debian packages) or rvm ruby (pick any version) most likely with rbench too (did not try).

The problem is fixed in newer version of the packages involved -- see the gem maintainers acknowledge there was a problem and fixed it.

All I want is that Conductor requires these fixed versions of packages because otherwise bundle install will fail on gcc with -Werror=format-security as seen in bug reports on github, mailing lists, Debian and Ubuntu.

There's no relation to any native package availability at all.

jguiditta:
Ok, I have no issue with this then, that seems pretty reasonable.  Tests already passed, so I'll merge this in.

PR #225, opened 2012-12-10
initial tim integration

jprovaznik:


n1zyy:
Attempting to import an image gives some confusing errors.

If I specify the remote image ID and give it a name, I get this flash error: "Image versions target images provider images base The requested image was not found on the Provider."

If I leave the name out (AFAIK, it's meant to be optional and populated from the provider if left blank), I get:

>    Name can't be blank
>    Name can't be blank
>    Image versions base image name can't be blank
>    Image versions target images image version base image name can't be blank
>    Image versions target images provider images target image image version base image name can't be blank
>    Image versions target images provider images base The requested image was not found on the Provider.

As discussed, this isn't necessarily a blocker for merging into the tim-stable branch (since it's still a WIP there), but wanted to note it here.

n1zyy:
Accessing just /tim/base_images/ throws an exception:

```
undefined local variable or method `image_row_partial' for #<#<Class:0x00000006a1c858>:0x00000007b1b5a0>
/home/a/aeolus/conductor/src/app/views/tim/base_images/_list.html.haml:2:in `block in _app_views_tim_base_images__list_html_haml__3885551483884135777_65599360'
```

It looks like it's because `image_row_partial` is passed in as a local variable when this is rendered from the Environments page, but not when accessing /tim/base_images/ directly.

src/app/controllers/application_controller.rb
```diff
@@ -46,7 +46,7 @@ class ApplicationController < ActionController::Base
   def handle_perm_error(error)
     if params[:return_from_permission_change]
       flash.now
-      redirect_to account_url
+      redirect_to main_app.account_url
```
jguiditta, line 49: main_app should not be needed _in_ the main app, this is a helper for accessing the main app from _within_ the engine.  Did you hit some issue here that required it to be in the host app?

n1zyy:
The answer may become apparent to me as I read further down, but are we going to break whatever API clients may exist (e.g., aeolus-cli)? Or is that okay at this point?

n1zyy:
Oh, re: image import -- it looks like that's what occurs for a non-existent image. With an image that does exist, I get `stack level too deep` instead. The entirety of the stack trace is `/home/a/.rvm/gems/ruby-1.9.3-p194@conductor/gems/activerecord-3.2.8/lib/active_record/transactions.rb:256` but it appears to come up when calling `@base_image.save` in the controller. I haven't followed it into the model yet.

FWIW, I have also upgraded to a newer version of Image Factory, though it appears to continue even after I downgrade to what I was using previously.

n1zyy:
I'm not sure if we did this previously or if it's always been missing, but I just uploaded a .jpg image as a template to build a new image, and was merrily taken to the edit_xml page with a textbox containing the image as text (!!), though it does complain that the syntax is invalid and decline to let me save it as-is.

jprovaznik:
re stack level too deep: sry, forgot to take care of this in patchset (this will have to be fixed on tim side in some way if they will stick with monkey patch). The problem is that currently tim integration patch uses "model_validations" branch of tim (not master) because there is an issue you describe above in Rails. But this model_validations branch doesn't load moneky patch ocrrectly, you have to add excplicit require somewhere temorary (e.g. initilizer).

jprovaznik:
re: uplaoding an image instead of template - I believe we did this before too, more sophisticated checking of uploaded file might be next-sprint task

src/app/controllers/application_controller.rb
jprovaznik, line 49: Yes, you will hit this issue even in ApplicationController, I suppose because Tim's ApplicationController inherits from this controller too.

jguiditta:
@n1zyy I would expect aeolus-cli will be broken initially, yes, the api is not even close to compatible with the factory/iwhd/1.0 version.  AFAIK, this will be addressed after conductor integration is complete/stable

src/app/controllers/application_controller.rb
jguiditta, line 49: hmm, that sounds 'less than ideal' at best.  If this is truly the case, then we may want to think about how and what we want to actually inherit in the engine's application controller (this can be a later task, not part of the initial integration effort)

n1zyy:
I left a bunch of nits, as you may have seen.

However, they're largely minor nits. The errors importing images, and the errors when accessing /tim/base_images/ directly are really the only things that I think are actively broken. The rest, I think we can merge into tim-stable and enhance from there. So I guess that's a conditional ACK, on fixing those two things.

Issue #207, opened 2012-11-30
API: Implement serialization of duration and timestamps according to XML Schema


jistr:
Implement helper methods that can be used across APIs. We should serialize time information according to XML Schema datatypes. This is what CIMI does as well.

Duration:
http://www.w3.org/TR/xmlschema-2/#duration

DateTime (timestamp):
http://www.w3.org/TR/xmlschema-2/#dateTime

PR #227, opened 2012-12-10
API: Time serializers

jistr:
Serializers for datetime and duration.

The duration serializer uses units upto days only, because months and years have varying length. 

Days have varying lengths too, but the differences are so small that I hope it does not matter for our purposes. If it should matter, we can always switch to using only units upto hours and still comply with XML Schema standard.

Addresses issue #207.


jistr:
Nooooooo :scream: 

Ruby 1.8 apparently doesn't have Float#to_r. Will fix and resubmit.

PR #228, opened 2012-12-10
Deployables index, third attempt

hacxman:
:-)

### DIFF
--- a/src/app/controllers/deployables_controller.rb
+++ b/src/app/controllers/deployables_controller.rb
@@ -25,12 +25,36 @@ class DeployablesController < ApplicationController
       @catalog = Catalog.find(params[:catalog_id])
       @deployables = @catalog.deployables
       @catalog_entries = @deployables.collect { |d| d.catalog_entries.first }
-    else
-      save_breadcrumb(deployables_path)
-      @deployables = Deployable.without_catalog.
+    elsif params[:pool_id].present?
+      # possible enumeration attack? as well as ^^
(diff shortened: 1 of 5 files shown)


src/app/controllers/deployables_controller.rb
```diff
@@ -25,12 +25,41 @@ def index
...
+      Deployable.transaction do 
+        Pool.find(params[:pool_id])
+        # only for it's side effect of raising exception in case of 404
+        catalogs = Catalog.find_all_by_pool_id(params[:pool_id])
+        deployables = catalogs.map { |_c| _c.deployables }.reduce([], &:+)
+
+        perm_deployables = Deployable.
+          list_for_user(current_session, current_user, Privilege::VIEW)
+
+        @deployables = deployables.to_set & perm_deployables.to_set
```
rwsu, line 43: Does the additions above need to be refactored? The code builds @deployables, but it is then not used in the response which I find confusing.

rwsu:
Please see code comment. 

Also there are a couple of extra whitespaces in the last patch. And at some point catalogs needs to be updated. The deployables that are listed there need an href.

hacxman:
I don't see why it needs to be refactored and well, I don't quite understand your comment on @deployables.

rwsu:
My mistake. I misread the if/else clauses. 

Conditional ack on whitespace issue.

hacxman:
New PR -> #255

PR #233, opened 2012-12-11
API: Time serializers v2

jistr:
[EDITED]

Fixed spec for Ruby 1.8.

Reading more of the XML Schema standard, durations composed of units upto days are considered totally ordered, so I guess XML Schema doesn't take leap seconds [1] into account. If we want to be less human-readable and avoid any ambiguity due to leap seconds, we can serialize durations using seconds as the only unit. I don't have a strong opinion here but kinda prefer serializing also with days, hours and minutes and not taking leap seconds into account (the way it is implemented now).

Replaces pull request #227, addresses issues #207.

[1] http://en.wikipedia.org/wiki/Leap_second

### DIFF
--- a/src/app/helpers/api_helper.rb
+++ b/src/app/helpers/api_helper.rb
@@ -23,6 +23,43 @@ module ApiHelper
       xmlschema_seconds_string(datetime.strftime('%S.%N'))
   end
 
+  # Serializes into http://www.w3.org/TR/xmlschema-2/#duration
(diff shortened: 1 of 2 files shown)


rwsu:
My preference would be to leave out days, hours, and minutes if their values are 0. For me it is more readable.

Looks good. Ack.

jistr:
Rebased and pushed, thanks for review :)

PR #235, opened 2012-12-11
BaseImages#show implementation

n1zyy:
This is a series of patches for the tim-stable branch that enables the image detail page.

### DIFF
--- a/src/app/decorators/controllers/tim/base_images_controller_decorator.rb
+++ b/src/app/decorators/controllers/tim/base_images_controller_decorator.rb
@@ -137,7 +137,7 @@ Tim::BaseImagesController.class_eval do
     @versions = @base_image.image_versions.order('created_at DESC')
     @latest_version = @versions.first # because we sorted above
     # @version is the specific image version we're viewing
-    @version = if params[:version]
+    @version = if params[:build]
       @versions.find{|v| v.id == params[:version]} || @latest_version
     else
       @latest_version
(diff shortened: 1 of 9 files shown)


n1zyy:
A few things I'm aware of that are non-perfect:
1. The link to push ProviderImages looks properly-constructed, but that method doesn't presently work. I believe this needs separate attention, though.
2. The "View Image Template XML" link does not work. I put a FIXME where I link to it. I also believe that this needs separate fixing.
3. I get a couple warnings (see below) about duplicate constants at startup. However, I cannot find them used anywhere else. Might this be something odd going on with initializers?

```
/home/a/aeolus/conductor/src/app/decorators/models/tim/provider_image_decorator.rb:25: warning: already initialized constant STATUS_COMPLETE
/home/a/aeolus/conductor/src/app/decorators/models/tim/provider_image_decorator.rb:26: warning: already initialized constant STATUS_IMPORTED
```

jprovaznik:
Hi Matt, I've added a few comments, but most of it are minor things. Only bigger thing which I'm missing is some way how to build newely create image ("rebuild" button), or was the plan to implement this in some oc upcoming patches?
Otherwise it looks good.

Btw, it seems this patchset breaks rspec tests.

n1zyy:
It seems like something is wrong with my test setup. I see that there are failures on Travis, though; I'll investigate.

I somehow overlooked the Rebuild feature. D'oh! I'll implement that, fix up what you mention, and figure out what's wrong with tests, then re-send. Thanks for the review!

n1zyy:
Updated and resent: https://github.com/aeolusproject/conductor/pull/256

PR #256, opened 2012-12-13
BaseImages#show, take 2

n1zyy:
This is an update to https://github.com/aeolusproject/conductor/pull/235
- The whitespace errors Jan found should be fixed
- `list_for_user` is used to filter provider accounts displayed
- I corrected the COMPLETE/COMPLETED constant value
- Imported images cannot be build/pushed
- I implemented a `build_all` method in the BaseImagesController. That's not 100% relevant to listing images, but as Jan points out, it's quite helpful to testing.
- As a bonus, I found and fixed a bug preventing users from selecting anything other than the most recent ImageVersion. I squashed it into the commit the bug was introduced in, so there's no trace that it ever happened :angel: 

The issue Jan noted with calling `imported?` on a ProviderImage and working our way up a few ancestors to check BaseImage's status is is ultimate a Tim bug/RFE, IMHO. I thought about redeclaring the method locally, but I'd rather just get it fixed in Tim.

### DIFF
--- a/src/app/decorators/controllers/tim/base_images_controller_decorator.rb
+++ b/src/app/decorators/controllers/tim/base_images_controller_decorator.rb
@@ -159,7 +159,12 @@ Tim::BaseImagesController.class_eval do
       _targetinfo = {
         :provider_type => provider_type,
         :target_image => target_image,
-        :provider_images => []
+        :provider_images => [],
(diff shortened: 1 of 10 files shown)


n1zyy:
I could not reproduce the test failures locally, BTW. Watching Travis eagerly...

PR #260, opened 2012-12-14
API: More info in deployments #show

jistr:


src/spec/requests/api/deployments_spec.rb
```diff
@@ -64,16 +66,14 @@ def check_deployment_xml_fields(xml_deployment, deployment)
     # xml_deployment.xpath("frontend_realm/@href").text.should == api_frontend_realm_url(deployment.frontend_realm)
     xml_deployment.xpath("uuid").text.should == deployment.uuid
     xml_deployment.xpath("scheduled_for_deletion").text.should == deployment.scheduled_for_deletion.to_s
+    xml_deployment.xpath("uptime_1st_instance").text.should ==
+      api_helper.xmlschema_absolute_duration(deployment.uptime_1st_instance)
+    xml_deployment.xpath("uptime_all").text.should ==
+      api_helper.xmlschema_absolute_duration(deployment.uptime_all)
+    xml_deployment.xpath("deployable_xml").inner_html.should == deployment.deployable_xml.to_s
```
rwsu, line 73: In mzatko's latest patch, the deployable xml is wrapped around a CDATA: https://github.com/aeolusproject/conductor/pull/270

We may need to discuss this again so everyone is in agreement on the approach.

rwsu:
Everything else looks good. 

Conditional ack pending outcome on deployables xml.

jistr:
Ah ok. If we wrap things in CDATA though, we should also handle nesting. Wrapping into CDATA without handling nesting (as in mzatko's patch) shouldn't be done. @hacxman please post your security concerns to the mailing list, so that we can move forward in deciding this (I see Richard mentioned you had security concerns about not using CDATA, but I can't find them).

jistr:
Oh and thanks for the review Richard :)

jistr:
It was concluded that we should not wrap in CDATA for now (and #284 got merged in), merging in this as well.

PR #262, opened 2012-12-14
Tim fixes1

jprovaznik:


n1zyy:
This looks good. Obviously we need to get the image template XML page working properly, but as you state, that's something for a future patch. ACK.

PR #263, opened 2012-12-14
Simplifying hardware_profiles form

ifarkas:


jcoufal:
Will look at this, but 1) failed Travis.

ifarkas:
@jcoufal, if you check the build log, you can see that not the tests failed but Travis failed to run the tests. I've run all the test suites locally and all pass for me. Does it fail for you locally?

jcoufal:
I did not test yet locally, it was just what I noticed at the first glance. Will look at it more properly later on...

jistr:
Yes, Travis sometimes fails recently, even when the code is ok. @ifarkas, if you want to force a rebuild, just do a minor change in the commit message of the last commit via `git commint --amend` and then force push to your pull request's branch. Changing the commit message will change the hash of the commit and that's a reason for Travis to kick off a new build.

ifarkas:
@jistr, Thanks for the suggestion. Apparently, closing the issue and reopening it also triggers a Travis rebuild. But both are just ugly hacks and belong to the toolset of Travis Driven Development. :-)

jcoufal:
General question:
- Do we want to allow to create HWP without filling at least one of the following fields: Memory, CPU, Storage, Architecture?

Some observations:
- Wrong input types for: Memory, CPU, Storage (expected is number type)
- Missing validations for:
  - Name (length)
  - Memory (size)
  - CPU (count)
  - Storage (size)
- I would expect some hints for fields (lentgh of the name, not negative memory, etc.)
- For CPU, add Note: Number of required CPUs.
  - (considering if help to the label wouldn't be helpful?)

That's all from UI (visual and functional) point of view. Needs to have a code check by someone else too.
Tests passed :)

n1zyy:
I just took a look at this, though it looks like Jarda's feedback is still relevant here.

I don't see a validation in the model on HWP name/label. The limit in the database appears to be 1024 characters, so it would take a real effort by the user to exceed the length, but Jarda's comment on validations made me look.

One nit -- if you click the 'New Hardware Profile' button form the edit page, it actually returns you to the show action of the HWP you are editing. I'm not sure whether it's a 'return' button that just has the wrong caption, or if it's meant to link to the 'new' action and the link is broken (in which case I'd argue we should just remove the button since it doesn't make sense).

ifarkas:
@n1zyy:

> One nit -- if you click the 'New Hardware Profile' button form the edit page, it actually returns you to the show action of the HWP you are editing. I'm not sure whether it's a 'return' button that just has the wrong caption, or if it's meant to link to the 'new' action and the link is broken (in which case I'd argue we should just remove the button since it doesn't make sense).

Nice catch! 

@jcoufal:

> - Missing validations for:
>   - Name (length)
>   - Memory (size)
>   - CPU (count)
>   - Storage (size)

The validations were defined on the HardwareProfileProperty class not on HardwareProfile. That might be the reason why you weren't able to find it. Although the numericality and greater_than options were already defined, I have added less_than options for all fields.

@n1zyy @jcoufal, Thanks for the review for both of you. All issues should now have been fixed.

jcoufal:
Fixed, looks ok. ACK (again, from the UI point of view).

In the future we would need to revisit hint messages, to be consistent. E.g. following part of message - it doesn't make sense to check HWP Name within Provider's scope:

"Must be a unique name within the set of Hardware Profiles belonging to the same Provider" is too long and . 

I will open new issue against it.

n1zyy:
I'm hitting conflicts trying to apply this. Would you rebase and resend or re-push?

ifarkas:
Rebased and pushed.

n1zyy:
ACK, assuming tests past. Your latest change fixed the nits that needed changing. Will merge after lunch if Travis says it's okay.

PR #265, opened 2012-12-17
Provider types api delete rm3884 - revision

petrblaho:
This one is revision of https://github.com/aeolusproject/conductor/pull/199

Adds permissions and checking for them when trying to delete Provider Type


martinpovolny:
we agreed on not fixing the redundant "each" as this will be used by a patch that will follow up shortly

but please fix the spaces after ( and before ) in src/spec/requests/api/provider_types_spec.rb

otherwise no problem

PR #266, opened 2012-12-17
Fixed redirects and build&push states

jprovaznik:


jprovaznik:
Going to send updated version of this patchset

PR #270, opened 2012-12-18
Deployables show, rebased

hacxman:


src/app/views/deployables/_show.xml.haml
```diff
@@ -0,0 +1,16 @@
+!!! XML
+%deployable{:id => deployable.id, :href => deployable_url(deployable)}
+  %name= deployable.name
+  %description= deployable.description
+  %xml
+    :cdata
+      "#{deployable.xml}"
+  %xml_filename= deployable.xml_filename
+  %owner{:id => deployable.owner.id, :href => user_url(deployable.owner)}
```
rwsu, line 9: Omit  owner for now. We'll add it back in once user resource has been implemented.

src/app/views/deployables/_show.xml.haml
```diff
@@ -0,0 +1,16 @@
+!!! XML
+%deployable{:id => deployable.id, :href => deployable_url(deployable)}
+  %name= deployable.name
+  %description= deployable.description
+  %xml
+    :cdata
```
rwsu, line 6: There was a discussion on aeolus-devel and there was agreement to embedded the xml instead of cdata. You brought up security/exploit concerns which are valid. We probably need to discuss this a bit more. There is a nesting issue with cdata which may or may not hit depending on how the deployable xml evolves: https://lists.fedorahosted.org/pipermail/aeolus-devel/2012-December/013516.html

src/app/views/deployables/_show.xml.haml
```diff
@@ -0,0 +1,16 @@
+!!! XML
+%deployable{:id => deployable.id, :href => deployable_url(deployable)}
```
rwsu, line 2: All calls to X_url() should be api_X_url()

rwsu:
routes.rb also needs to be updated.

hacxman:
new PR is: #284, so closing this one

PR #271, opened 2012-12-18
Deployables delete, rebased

hacxman:


src/app/controllers/deployables_controller.rb
```diff
@@ -308,6 +308,17 @@ def destroy
           redirect_to deployables_path
         end
       end
+
+      format.xml do
+        if deployable.destroyed?
+          render :text => '', :status => :no_content
+        else
+          # TODO: this is wrong'n'ugly, please come with smtg better
```
rwsu, line 316: To be consistent with other controllers, we should return code 500. 

raise Aeolus::Conductor::API::Error.new(500, deployablet.errors.full_messages.join(', '))

rwsu:
routes.rb also needs to be updated.

hacxman:
PR: #278

PR #272, opened 2012-12-18
API Entrypoint: add Deployables

hacxman:


rwsu:
entrypoint_controller_spec.rb needs to be updated.

Issue #171, opened 2012-11-16
Instances api - list, show, entrypoint


rwsu:
Initial implementation of the instances api.

PR #273, opened 2012-12-19
API: Instances index, show, and entrypoint 

rwsu:
Replaces #171.

Removed fields from show that are not ready to be displayed.

Updated date fields to new format.

Reference to deployment is now included in show.

Full resource representation is shown in top level collections.

### DIFF
--- a/src/app/controllers/instances_controller.rb
+++ b/src/app/controllers/instances_controller.rb
@@ -30,6 +30,7 @@ class InstancesController < ApplicationController
       format.html
       format.js { render :partial => 'list' }
       format.json { render :json => @instances.map{ |instance| view_context.instance_for_mustache(instance) } }
+      format.xml { render :partial => 'list.xml' }
     end
   end
 
@@ -55,6 +56,7 @@ class InstancesController < ApplicationController
         render :partial => @details_tab[:view] and return
       end
       format.json { render :json => @instance }
+      format.xml { render :show, :locals => { :instance => @instance } }
     end
   end
 
--- a/src/app/controllers/instances_controller.rb
+++ b/src/app/controllers/instances_controller.rb
@@ -30,7 +30,7 @@ class InstancesController < ApplicationController
       format.html
       format.js { render :partial => 'list' }
       format.json { render :json => @instances.map{ |instance| view_context.instance_for_mustache(instance) } }
-      format.xml { render :partial => 'list.xml' }
+      format.xml { @deployment = @instances.first.deployment if params[:deployment_id] }
     end
   end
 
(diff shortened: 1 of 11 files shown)


rwsu:
Here is a stack trace of the build error. 

malformed format string - %:
/home/rwsu/conductor/src/vendor/ruby/1.8/gems/i18n-0.6.0/lib/i18n/core_ext/string/interpolate.rb:101:in `interpolate_without_ruby_19_syntax'
 /home/rwsu/conductor/src/vendor/ruby/1.8/gems/i18n-0.6.0/lib/i18n/core_ext/string/interpolate.rb:101:in`send'
 /home/rwsu/conductor/src/vendor/ruby/1.8/gems/i18n-0.6.0/lib/i18n/core_ext/string/interpolate.rb:101:in `%'
 /home/rwsu/conductor/src/app/helpers/api_helper.rb:22:in`xmlschema_datetime'
 /home/rwsu/conductor/src/app/views/instances/_detail.xml.haml:12:in `_app_views_instances__detail_xml_haml__477859814_70277688006900'

jistr:
Found the problem. The helper method is tested to work with DateTime, but I suppose Ruby classes aren't good enough for Rails :) I just looked at updated_at attribute, and it has a type ActiveSupport::TimeWithZone and their implementation of strftime doesn't seem to work in a standard way on Ruby 1.8. :rage4: Fortunately, they provide a to_datetime method on this thing, so this should fix it.

Don't change your code, I'm going to harden the helper (make it convert the parameter to DateTime) and submit a pull request for it. Then please review/merge it and rebase your pull request on top of that, and I hope it should be ok ;)

jistr:
I hope that pull request #274 will fix it. Couldn't test it live on my machine because rbenv cannot compile Ruby 1.8 on the newer versions of GCC. Please ping me if the problem persists.

src/app/controllers/instances_controller.rb
```diff
@@ -55,6 +56,7 @@ def show
         render :partial => @details_tab[:view] and return
       end
       format.json { render :json => @instance }
+      format.xml { render :show, :locals => { :instance => @instance } }
```
jistr, line 59: Not totally sure, but I think the final consensus was that we shouldn't use :locals for passing from controllers to views when possible, and should use :locals for passing params to partials when possible [1]. It's a minor issue though.

https://lists.fedorahosted.org/pipermail/aeolus-devel/2012-August/thread.html#12054

jistr:
Looks good to me and works for me. Conditional ACK with regard to the Ruby 1.8 thing.

PR #274, opened 2012-12-19
API: Make time helpers more defensive

jistr:
Check for nils, convert datetime parameter to the true datetime (to prevent
errors when using ActiveSupport::TimeWithZone on Ruby 1.8 [1]).

[1] https://github.com/aeolusproject/conductor/pull/273#issuecomment-11516852


rwsu:
Ack.

PR #282, opened 2012-12-20
fix cost formating in deployment and instance properties

martinpovolny:


jistr:
Looks good and I tried it manually, ACK and merging.

PR #284, opened 2012-12-20
Deployables show rebase, take 2

hacxman:
rebased after delete was merged


rwsu:
Ack. We'll need to revist the embedded xml to see if it needs to be namespaced.

PR #297, opened 2013-01-02
a small step on unifying the (error) logging

martinpovolny:
this patch unifies the backtrace logging

please note that I am not trying to fix what is logged and when

this just makes all the places where backtraces are logged look the same


petrblaho:
ACKed and merged.

Issue #261, opened 2012-12-14
removing deployable from catalog broken


martinpovolny:
I am e.g. at /catalogs/1/deployables/2 

then I click the small cross (input class="remove_catalog")

then I get to:

/catalog_entries/4

with an errror message:

   No route matches {:action=>"show", :controller=>"deployables", :catalog_id=>#<Catalog id: 3, pool_id:
   3, name: "a catalog for rhevm", created_at: "2012-10-31 09:42:54", updated_at: "2012-10-31 09:42:54",
   pool_family_id: 3>, :id=>#<Deployable id: 4, name: "rhevm_1", description: "", xml: "<?xml 
   version=\"1.0\"?>\r\n<deployable version=\"1.0\" na...", xml_filename: "just an image from rhevm #1",
   owner_id: 2, pool_family_id: 3>}

martinpovolny:
then if I click "back" and click the small cross again, I get to

/catalog_entries

saying: Unknown action

The action 'index' could not be found for CatalogEntriesController

whole this workflow is missing a test case

PR #300, opened 2013-01-03
fix issue #261

martinpovolny:
just a quick fix for issue #261


jzigmund:
It's fixed the bug as we expect, ack & merge.

PR #301, opened 2013-01-04
fix secret_token initialization

martinpovolny:
we need to fix aeolus-configure to match this change


tomassedovic:
Other than the copyright header nit, this is great. Works swell and fixes a pretty embarrassing security issue. ACK.

Thanks!

PR #304, opened 2013-01-07
Link to http://www.aeolusproject.org instead of https variant

iNecas:
https://www.aeolusproject.org is down.


n1zyy:
Nice catch, thanks! (We moved the site to GitHub pages, which doesn't do HTTPS.)

ACK, merging.

PR #316, opened 2013-01-07
Edit src/config/compass.rb to match newest compass reqs

eggmaster:
This removes three lines from src/config/compass.rb because they
are incompatible with the newest upstream version of rubygem-compass.


bmclaughlin:
Compass issue confirmed on F16, F17 and Travis build passes as well.  Merging.

PR #317, opened 2013-01-08
Instance API #show should check provider account is not null

rwsu:
Instances created in non-test environment may have null as its
provider_account_id.


jistr:
ACK

Issue #246, opened 2012-12-13
Bring the asset pipeline and alchemy gem into conductor


aathomas:
- remove converge-ui submodule and symlinks
- bring in alchemy gem
- assets pipeline in conductor - configuration, manifest files
- rpmbuild changes to deal with precompiling the assets

PR #318, opened 2013-01-08
#246 Asset pipeline, Alchemy

jtomasek:
This pull request implements https://github.com/aeolusproject/conductor/issues/246
- assets pipeline
- alchemy gem instead of converge-ui git submodule
- prepares Conductor rpm building for Fedora 18


eggmaster:
See comment on c785c98. This pull request is largely working for me, but I think there are some corrections needed in the list of gem dependencies.

PR #327, opened 2013-01-09
Rebase fixes

jprovaznik:


n1zyy:
All tests green on 1.9.

Failures on 1.8 are known, and introduced by Deltacloud. Going to merge this; we should work with them to get these fixed before merging this to master, but this is good for tim-stable. (And besides, this patch isn't introducing any new failures; they were already there.)

Code passes visual check, too.

PR #329, opened 2013-01-14
Fixed text_field html id

jprovaznik:


jprovaznik:
this is minor nit, self-ack

PR #344, opened 2013-01-16
Prevent users from modifying their own quota

n1zyy:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=871136


bmclaughlin:
ACK, removes the option from unprivileged users.

PR #350, opened 2013-01-17
Openstack -> OpenStack string cleanup

### LINKED ISSUE #333: 'Openstack' ProviderType should be 'OpenStack'
This seems minor and nitpicky, but we need to get it right -- OpenStack is properly written with the 'S' capitalized. The ProviderType included in our db/seeds.rb does not match this usage and should be updated.

n1zyy:
We improperly capitalized 'Openstack' in a few places.

Closes https://github.com/aeolusproject/conductor/issues/333

(Nevermind that I pushed this feature branch to the aeolusproject repo by mistake -- when you merge it, it will ask if you want to delete the branch I pushed, which you can go ahead and do.)

### DIFF
--- a/src/config/locales/en.yml
+++ b/src/config/locales/en.yml
@@ -1127,7 +1127,7 @@ en:
         rhevm: RHEV-M API_PROVIDER
         vsphere: vSphere API hostname
         rackspace: Rackspace API_PROVIDER
-        openstack: Openstack API_ENTRYPOINT
+        openstack: OpenStack API Endpoint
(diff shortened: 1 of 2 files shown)


n1zyy:
Actually, this should not be merged until https://github.com/aeolusproject/conductor/pull/352 is taken care of. Otherwise, if someone added 'Openstack', checking for 'OpenStack' will not find it, causing the OpenStack provider to disappear.

jprovaznik:
Works fine. Tests actually pass. Merging.

PR #351, opened 2013-01-17
Removes 'show me the page' test step

n1zyy:
This was popping open elinks while running tests. I don't think
this should be in the repo, just added locally for debugging.


jguiditta:
I think this qualifies as 'safe to just push without a pull request  :)

PR #352, opened 2013-01-17
Removes available_providers filter in ProvidersController

n1zyy:
This:

a.) Served no constructive purpose. It was ostensibly to allow
 upstream to support everything, but product to only support
 fully-tested providers. Except the file was never changed.

b.) Does not belong in a controller.

c.) Breaks if you send a patch like my 'Openstack' -> 'OpenStack'
 one, unless I were to check for both strings, which would be
 a ridiculous mess.

### DIFF
--- a/src/app/controllers/providers_controller.rb
+++ b/src/app/controllers/providers_controller.rb
@@ -302,8 +302,7 @@ class ProvidersController < ApplicationController
   end
 
   def load_providers_types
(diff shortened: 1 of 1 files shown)


jprovaznik:
Travis seems to be broken, but tests actually pass. I've created an issue #353 which is related to this.
Going to merge.

PR #356, opened 2013-01-18
Fixes glance_url to use a string, not a text field

### LINKED ISSUE #335: Glance URL field is enormous
Adding a new provider account for OpenStack, the text field seems like an enormous textarea. It should be a normal input like all of the other fields.

n1zyy:
This updates db/seeds.rb only, without a migration. But since we're
updating before the release, there should be need for one.

It turns out that :type => 'text' gives you a textarea...

Closes https://github.com/aeolusproject/conductor/issues/335

### DIFF
--- a/src/db/seeds.rb
+++ b/src/db/seeds.rb
@@ -218,6 +218,6 @@ if CredentialDefinition.all.empty?
 
   # For OpenStack provider type:
   if openstack = ProviderType.find_by_deltacloud_driver('openstack')
-    CredentialDefinition.create!(:name => 'glance_url', :label => 'glance_url', :input_type => 'text', :provider_type_id => openstack.id)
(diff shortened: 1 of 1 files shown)


src/db/seeds.rb
```diff
@@ -218,6 +218,6 @@
 
   # For OpenStack provider type:
   if openstack = ProviderType.find_by_deltacloud_driver('openstack')
-    CredentialDefinition.create!(:name => 'glance_url', :label => 'glance_url', :input_type => 'text', :provider_type_id => openstack.id)
+    CredentialDefinition.create!(:name => 'glance_url', :label => 'glance_url', :input_type => 'string', :provider_type_id => openstack.id)
```
jguiditta, line 221: so, are the ui and api unaffected by this change?

PR #357, opened 2013-01-18
Deployables create improve test coverage

hacxman:
some (2) of those fails though, I believe it is due to current implementation being incorrect
'fixes' #322


n1zyy:
@hacxman - This looks like it needs to be rebased against master.

hacxman:
rebased & fixed whitespace error

martinpovolny:
@hacxman so to merge this we would probably like to fix the implementation in the same pull-request

or do you have different idea?

martinpovolny:
nice tests, code finally fixed ;-)

ack and merging

PR #358, opened 2013-01-18
Delete a bunch of old images

n1zyy:
These are only images that I was >99% sure we did not need. There
are probably still some random graphics lingering around, but I
think I got the bulk of the obviously-orphaned ones.


n1zyy:
I didn't set out to do this -- eggs asked about some long-outdated image that rpmdiff was complaining about, and in verifying it was obsolete, I found a bunch of wild stuff.

Incidentally, some of these graphics are good stuff -- but not used in our app or with obvious near-term use IMHO.

hacxman:
ack

PR #277, opened 2013-01-18
Tim integration - remove iwhd from aeolus setup and rpms

jeckersb:


jprovaznik:
Tests actually pass, looks good, merging.

PR #360, opened 2013-01-21
fix cost est. for instance in some cases (rhevm)

### LINKED ISSUE #359: "Cost Estimate" for rhevm instance is not getting calculated
"Cost Estimate" for rhevm instance not getting calculated, and also observed that when the instance status changes from vanished--> running the "cost estimate " was displaying Negative value.

martinpovolny:
fixes: #359

this is a minor fix

more elaborate redo of up-time calculation for Instance and then cost estimate will be needed


jprovaznik:
Merging this - this is hotfix for most common error on RHEVM, Long-term fix will be part of #361 issue.

PR #363, opened 2013-01-21
BaseImages#show shows provider accounts whether or not images are built

### LINKED ISSUE #306: BaseImages#show should show provider accounts when none are built
_This refers to the tim-stable branch, not yet merged to master._

When defining a new image not yet built, the BaseImage show page shows no providers or provider accounts, only the "Build All" button. They show up once you click Build All.

This is a design flaw on my part -- we should be showing the list whether or not you have any builds.

n1zyy:
This fixes a bug where the BaseImage detail page would only show the
available provider accounts if they had an image built.

Fixes https://github.com/aeolusproject/conductor/issues/306

### DIFF
--- a/src/app/decorators/controllers/tim/base_images_controller_decorator.rb
+++ b/src/app/decorators/controllers/tim/base_images_controller_decorator.rb
@@ -176,15 +176,15 @@ Tim::BaseImagesController.class_eval do
     end
   end
 
+  # This sets up @targets, which is a big hash with information on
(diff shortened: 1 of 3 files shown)


n1zyy:
The diff formatting doesn't make it quite clear, but I ended up just rewriting a few of the involved methods, to do things a little more cleanly. I also changed a few variable names that inappropriately referenced images, when `@targets` is much more than images.

PR #364, opened 2013-01-22
Remove iwhd from settings.yml

cwolferh:
No longer used now that conductor uses tim.


jeckersb:
I did this in #277 along with removing every other reference to iwhd (if you want to review it, that would be great!)

cwolferh:
Sticking with the prior pull request, closing this one.

PR #366, opened 2013-01-22
Providers page refactor: switch providers#edit with providers#show

ifarkas:
Until now, we used the providers#show page to display the providers form. This pull request takes care switching the content of the show and edit action. It also changes a few things to be more consistent with other parts of Conductor.


n1zyy:
ACK. I had actually grown so used to being taken to the edit page that this change was jarring, but it does make it consistent with the rest of the app (and with normal web patterns). Merging.

PR #378, opened 2013-01-25
Added snapshot support for image build&push

### LINKED ISSUE #276: Tim integration - enable snapshot mode for image building

jprovaznik:
Now user can set in config/settings.yml file for if she wants use
snasphot mode for providers for which imagefactory supports snapshots.
Currently these are: ec2, openstack, mock

This patch also removes the old annoying restriction when provider name
was used for finding realm on imagefactory side.

Fixes #276


jprovaznik:
Note that a fix on imagefactory side is required too: https://github.com/aeolusproject/imagefactory/pull/227

n1zyy:
This looks good and tests OK. ACK; merging.

PR #391, opened 2013-01-28
remove all foreign keys until rails is fixed

martinpovolny:
remove all foreign keys from database until Rails is released with this fixed:  
https://github.com/rails/rails/pull/8548                                        

keeping the FKs causes dev-tools to fail as described in the Rails issue


jguiditta:
Ok, so there are 2 fails on specs+1.8, but as martin assures me they are a separate fix, merging in this request, as it fixes dev-tools install

PR #412, opened 2013-02-21
DeployableMatching refactor

ifarkas:
Refactor deployable matching logic to a separate lib


jprovaznik:
Hope I didn't miss something (it's quite big), but this looks good and works fine. Merging. Great job!

PR #414, opened 2013-02-21
Gettext last

jzigmund:
Last patchset contains t-keys conversion with interpolation and count variable.

### DIFF
--- a/src/app/models/instance.rb
+++ b/src/app/models/instance.rb
@@ -120,6 +120,16 @@ class Instance < ActiveRecord::Base
   STATE_ERROR          = "error"
   STATE_VANISHED       = "vanished"
 
+  N_('new')
+  N_('pending')
+  N_('running')
+  N_('shutting_down')
+  N_('stopped')
+  N_('stopping')
+  N_('create_failed')
+  N_('error')
+  N_('vanished')
+
   STATES = [STATE_NEW, STATE_PENDING, STATE_RUNNING,
              STATE_SHUTTING_DOWN, STATE_STOPPED, STATE_CREATE_FAILED,
              STATE_ERROR, STATE_VANISHED]
@@ -603,7 +613,9 @@ class Instance < ActiveRecord::Base
 
   def do_operation(user, operation)
     task = self.queue_action(user, operation)
-    raise I18n.t("instances.errors.#{operation}_invalid_action") unless task
+    unless task
+      raise operation == 'stop' ? _('Stop is an invalid action.') : _('Reboot is an invalid action.')
+    end
     Taskomatic.send("#{operation}_instance", task)
   end
 
--- a/src/app/models/provider_account.rb
+++ b/src/app/models/provider_account.rb
@@ -76,6 +76,14 @@ class ProviderAccount < ActiveRecord::Base
   # will result in an exception.
   attr_protected :credentials_hash
 
+  N_('Username')
+  N_('Password')
+  N_('Access Key')
+  N_('Secret Access Key')
+  N_('Account Number')
+  N_('Key')
+  N_('Certificate')
+
   # Validations
   validates :label, :presence => true,
                     :uniqueness => true,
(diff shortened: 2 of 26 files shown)


src/app/models/deployable.rb
```diff
@@ -54,6 +54,14 @@ class << self
 
   PRESET_FILTERS_OPTIONS = []
 
+  BUILD_STATUS = {
+      :not_build => _("Images are not Built"),
```
ifarkas: 'Built' should not be capitalized.

src/app/models/deployable.rb
```diff
@@ -54,6 +54,14 @@ class << self
...
+      :building => _("Images are being built."),
+      :pushing => _("Images are being pushed."),
+      :not_pushed => _("Some of the images are not pushed"),
+      :pushed => _("All Images are pushed and recent.")
+  }
+
```
ifarkas: A problem with this solution is that the translations are being cached on the Deployable model. It means that even if you switch locale during runtime, it won't be picked up in production environment.

src/app/models/instance.rb
```diff
@@ -120,6 +120,16 @@ class << self
...
+  N_('new')
+  N_('pending')
+  N_('running')
+  N_('shutting_down')
+  N_('stopped')
+  N_('stopping')
+  N_('create_failed')
+  N_('error')
+  N_('vanished')
+
```
ifarkas, line 132: I don't like the idea of adding random Gettext function calls to the model which have nothing to do with the functionality of model. Since our models are way too oversized, how about adding these strings to locale/unfound_translations.rb or similar which is also recommended by the readme of gettext_i18n_rails?

src/app/models/deployment.rb
```diff
@@ -502,6 +502,31 @@ def events_of_deployment_and_instances
...
+        _('All Instances are in failed state')
+      when STATE_ROLLBACK_IN_PROGRESS
+        _('Launch failed, rollback is in progress')
+      when STATE_ROLLBACK_COMPLETE
+        _('Rollback successfully completed')
+      when STATE_ROLLBACK_FAILED
+        _('Rollback failed, re-launch terminated')
+    end
+  end
+
```
ifarkas: Same here, this doesn't belong to model layer.

src/app/models/provider_account.rb
```diff
@@ -78,6 +78,14 @@ class << self
...
   attr_protected :credentials_hash
 
+  N_('Username')
+  N_('Password')
+  N_('Access Key')
+  N_('Secret Access Key')
+  N_('Account Number')
+  N_('Key')
+  N_('Certificate')
+
```
ifarkas, line 86: Same here, this doesn't belong to model layer.

src/app/models/user_group.rb
```diff
@@ -26,6 +26,9 @@ class << self
   MEMBERSHIP_SOURCE_LOCAL = "local"
   MEMBERSHIP_SOURCES = [MEMBERSHIP_SOURCE_LOCAL, MEMBERSHIP_SOURCE_LDAP]
 
+  N_("LDAP")
+  N_("local")
+
```
ifarkas, line 31: Same here, this doesn't belong to model layer.

jzigmund:
Pull request was updated, issues are fixed + configuration has been moved to initializers/fast_gettext.rb (according docs for fast_gettext gem)

ifarkas:
I think you might forget to delete the Deployable::BUILD_STATUS hash. The functionality seems redundant since it's also implemented in DeployableHelper.

jzigmund:
Updated and deleted mentioned hash

ifarkas:
I have found one more thing. In current upstream/master we can change the locale as usual:

``` ruby
FastGettext.locale # => "en" 
FastGettext.locale = :de # => :de 
FastGettext.locale # => "de" 
```

...but in your branch it is broken:

``` ruby
FastGettext.locale # => "en" 
FastGettext.locale = :de # => :de 
FastGettext.locale # => "en" 
FastGettext.locale = 'de' # => "de" 
FastGettext.locale # => "en" 
```

I am not really sure why this is happening, tried to debug it but can't find what is causing this issue. Maybe because of the changes in application.rb?

ifarkas:
As we discussed with Jozef, only locales listed in FastGettext.default_available_locales can be enabled with this change. It can be edited in config/initializers/fast_gettext.rb.

ACK

PR #415, opened 2013-02-22
Remove as_json methods from models

ifarkas:
These were used with backbone calls before we integrated with Mustache


jzigmund:
ACK

PR #417, opened 2013-02-22
Remove image caching from Deployable model

ifarkas:
The caching was added because the image details were queried from ImageWarehouse over HTTP. With TIM, it's all stored in DB, so no need for this kind of caching.


jzigmund:
ACK.

PR #418, opened 2013-02-25
Fix Ruby 2.0 compatibility issues

ifarkas:


src/Gemfile
```diff
@@ -1,4 +1,4 @@
-source :rubygems
```
jguiditta, line 1: does this really break ruby 2.0?  The bundler docs show them as equivalent: http://gembundler.com/v1.2/gemfile.html

jguiditta, line 1: ahh, just saw your commit msg, that is too bad bundler doesn't default to https

PR #14, opened 2012-08-22
Minor coding style fixes across models

mfojtik:
- Removed unused variables
- Removed unnecessary return statements
- Fixed formatting on Hashes

PR #422, opened 2013-02-25
code style cleanup in models

martinpovolny:
code styles fixes similar to #14 but not breaking tests and above current master


ifarkas:
ACK

PR #425, opened 2013-02-28
use unreleased dc-client version to prevent problems with openstack

martinpovolny:
need to get an unreleased version of dc-client to get openstack working


jprovaznik:
merging this, according to feedback this doesn't break anything else

PR #428, opened 2013-03-01
Enable slow query warnings in development environment

ifarkas:
This feature was introduced in Rails 3.2 and is enabled by default in new projects.


martinpovolny:
ook, so you wanna more information in the logs, have it

the most valid argument from my point of view is that is't the default in rails and it's good not to deviate where it's not needed

ifarkas:
> the most valid argument from my point of view is that is't the default in rails and it's good not to deviate where it's not needed

Yeah, I agree. I would have not created this pull request if it is not the default in Rails 3.2

PR #430, opened 2013-03-12
gettext conversion for role UI and role names

sseago:


jzigmund:
Please rebase the patch. Except that one nit I'm giving ACK to the patch.

PR #432, opened 2013-03-14
Refactor Provider Account priority 

ifarkas:
This patchset includes the following changes:
- Provider Account priority is removed
- the score ranking in provider selection previously was based on the Provider Account priority which caused confusion in many situations. With this patch the ranking goes from -100 (lowest) to +100 (highest) and nil value worth 0.
- Weighting strategy is added


martinpovolny:
I would shorten this function, too verbose for my feeling of "pretty"

``` ruby
+Then /^I should see score "(.*?)" for provider account with label "(.*?)"$/ do |score, provider_account_label|
+  found = false
+  all("tr").each do |tr|
+    if tr.has_content?(provider_account_label) && tr.has_content?(score)
+      found = true
+    end
+  end
+  assert found
+end
```

step #1:

``` ruby
found = false                                                                   
all("tr").each do |tr|                                                          
  found ||= tr.has_content?(provider_account_label) && tr.has_content?(score)   
end                                                                             
assert found     
```

step #2:

``` ruby
assert all("tr").detect { |tr| tr.has_content?(provider_account_label) && tr.has_content?(score) }
```

over all looks good, works, tests pass

ACK and merging it, just consider revisiting the function in question, please

ifarkas:
@martinpovolny, sure, I will create a new pull request fixing that. Thanks for the review!

PR #436, opened 2013-03-21
Remove Ruby 1.8 from Travis build matrix

ifarkas:
Ruby 1.8 EOL is on June 2013. Let's not consume spare Travis cycles if we don't need to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - 1.8.7
   - 1.9.3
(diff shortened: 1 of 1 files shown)


ifarkas:
As far as I know, we are not going to release any new version of Conductor targeting Ruby 1.8. If I am wrong or you have any objection against it, please comment here.

martinpovolny:
+1 for this suggestion from me

martinpovolny:
no objections, merging

PR #438, opened 2013-04-04
Alberich integration

sseago:
Removed local permissions code and use the Alberich engine instead.

A couple notes:
1) Filter/search UI for the permissions tab on object pages doesn't currently work due to a routing/redirect issue, this should be resolved at some point, but it's not a huge priority. Filter works fine on the permissions list on the user/group profile pages
2) in order to use this, you need to build and install the alberich gem including the content in the conductor-integration pull request (https://github.com/aeolus-incubator/alberich/pull/5 )


jguiditta:
This looks reasonable to me.  Better to get it in and smooth out issue later as needed if they pop up imo, so I'll merge it in

